PR #20, opened 2020-04-21
Add battery disclaimer

jacobgkau:
After today's discussion, this PR:

- Moves the main battery section to the bottom of the repairs page.
- Adds the approved disclaimer text to this section.
- Updates the difficulty level to "high" since we are not going to recommend customers do this themselves.


jacobgkau:
Do we want to add a horizontal rule and make "non user-serviceable tasks" more separated, or is this sufficient?

jackpot51:
This looks great to me. Since having the disclaimer is a high priority I have merged it but feel free to make a PR with any other improvements

PR #21, opened 2020-06-05
Add Gazelle 15

jacobgkau:
This PR adds the specifications, internal overview, external overview, and service manual for the Gazelle 15.

The port photos are currently labeled incorrectly:
- The USB Type-C port is USB 3.2 on both graphics versions.
- The USB Type-C port only supports DisplayPort output on the 1660 Ti.

I don't have the SVG files to update the port diagrams. We'll either need to wait for those to get updated before merging this, or merge first and update them later.


src/models/README.md
```diff
@@ -3,5 +3,6 @@
 System76 provides laptops, desktops and servers with Linux pre-installed.
 This documentation includes the following System76 models:
 - [Adder WS (addw1)](addw1/README.md)
+- [Gazelle (gaze15)](addw1/README.md)
```
jackpot51, line 6: This currently points to the addw1 docs

jacobgkau: Fixed.

jackpot51:
Can you squash this into one commit?

PR #26, opened 2020-08-13
Correct M.2 drive specs on gaze15

leviport:


src/models/gaze15/repairs.md
```diff
@@ -90,7 +90,7 @@ RAM acts as temporary storage for your computer. More RAM generally provides bet
 
 ## Replacing an M.2/NVMe SSD:
 
-This model supports up to two M.2 SSDs. Both M.2 slots are size 2280, and both slots support SATA III or PCIe NVMe Generation 3.
+This model supports up to two M.2 SSDs. Both M.2 slots are size 2280. The top slot (shown with the yellow box in the image below) supports either SATA III or PCIe NVMe Generation 3, and the bottom slot (orange box) supports only NVMe Generation 3.
```
jacobgkau, line 93: I know it gets long-winded, but you called it "PCIe NVMe Generation 3" the first time and just "NVMe Generation 3" the second time. If you're opposed to using the full name the second time due to length, that's fine, just wanted to point it out because someone will probably end up asking us what the difference is.

leviport: Good catch, I'll change it. Thanks!

jacobgkau:
Do you want to correct the tech specs on the main overview/readme at the same time, while we have the PR open? Just take away "/SATA" from one of the slots. After that, this should be good to go.

leviport:
Thanks, I forgot it was mentioned there too. Latest commit addresses that.

jacobgkau:
I actually meant this one.

PR #34, opened 2020-09-30
Correct information about SATA/NVMe compatibility

jacobgkau:
The serw12 only supports SATA in one of its slots, not both. Additionally, the slot that is NVMe-only is limited to PCIe Gen 2 (testing shows write speeds are unaffected but read speeds are about 50% slower.)

The reason I refer to the slower NVMe-only slot as "Slot 1" is because that's what is physically printed on the board, and I believe it would lead to confusion if we referred to them opposite what they are physically labeled.

The reason I state that Gen 3 drives will work at slower speeds in the Gen 2 slot is not to call attention to the slower speeds (although that's important too), but rather to alleviate concern about Gen 3 drive compatibility.


leviport (approved):
Well that's a bummer. Nice find though.

PR #35, opened 2020-10-08
Correct bottom panel screw count and placement

jacobgkau:
This fixes two issues:

1. Updates the bottom panel photo to show the correct screw placement (I didn't have access to a gaze15 when initially creating this section, which is why gaze14 photos were re-used.)

2. Corrects the screw counts to show a total of 17 instead of 18 (this mistake was imported from the gaze14 service manual, which shows 17 screws in its photo but gives numbers adding up to 18.)


ahoneybun (approved):
LGTM! Thank you for updating this!

PR #38, opened 2020-10-15
Add Meer5 to tech-docs

leviport:


src/models/meer5/README.md
```diff
@@ -0,0 +1,37 @@
+# Meerkat (meer5)
+
+- [External Overview](./external-overview.md)
+- [Internal Overview](./internal-overview.md)
+- [Parts & Repairs](./repairs.md)
+
+![Short Meerkat](./img/meer5-short.png)
+
+The System76 Thelio Massive is a desktop with the following specifications:
```
ahoneybun, line 9: Thelio Massive? Are these specs correct?

ahoneybun (changes requested):

src/models/meer5/README.md
leviport: Hey, meerkats can dream, can't they? 

Thanks for catching that. I was copying chunks of Massive's docs to use as a template and forgot to update that. The specs are correct.

ahoneybun:
When I run it live using `mdbook serve` I don't see the meer5. I can see the files being loaded but the .html files are not created for that model. Do you see that @leviport ?

leviport:
> When I run it live using mdbook serve I don't see the meer5. I can see the files being loaded but the .html files are not created for that model. Do you see that @leviport ?

Whoops, I think I know what the issue is. Just a moment.

ahoneybun (approved):
LGTM!

PR #39, opened 2020-10-15
Correct image file extension/format to match link

jacobgkau:
https://github.com/system76/tech-docs/pull/38 added the M.2 drive image in .png format, but linked to it in .jpg format, so it's currently not showing up (see [here](https://tech-docs.system76.com/models/meer5/repairs.html#replacing-the-m2-drive)). This PR updates the image to be in .jpg format, so it's smaller and shows up properly.

Also removed the words "switch off the power supply" since there is no switch on the Meerkat, and made a couple of very small grammatical improvements.


leviport (approved):
Thanks!

PR #41, opened 2020-11-02
Update OS versions

leviport:


ahoneybun:
Wouldn't we want to do this with the Massive as well correct?

leviport:
OS versions aren't mentioned in the Massive guides, are they?

ahoneybun:
Should we just have it as Pop!_OS and Ubuntu on both or remove it from the Thelio Major?

leviport:
I've removed the OS sections completely. I think this page will need more lovin' eventually, but this at least gets rid of the inaccurate information as a quick-fix.

ahoneybun (approved):
LGTM thank you!

PR #49, opened 2020-12-17
lemp10: correct part number & link for soldered RAM

leviport:
The linked part number was for a 16GB 2666MHz chip, so I changed it to this 8GB 3200MHz chip. It looks like [there was another possibilitiy](https://www.samsung.com/semiconductor/dram/ddr4/K4A8G085WC-BIWE/) so it might be worth verifying that I got it right. It's definitely not high priority though, since this corrects the size and speed discrepancy.

### DIFF
--- a/src/models/lemp10/README.md
+++ b/src/models/lemp10/README.md
@@ -22,7 +22,7 @@ The System76 Lemur Pro is a laptop with the following specifications:
     - eDP 14.1" 1920x1080 LCD
     - HDMI and DisplayPort 1.2 over USB-C
 - Memory
-    - Channel 0: 8-GB on-board DDR4 [Samsung K4A8G085WC-BCWE](https://www.samsung.com/semiconductor/dram/ddr4/K4A8G085WC-BCWE/)
+    - Channel 0: 8-GB on-board DDR4 [K4AAG165WA-BCWE](https://www.samsung.com/semiconductor/dram/ddr4/K4AAG165WA-BCWE/) X 8
     - Channel 1: 8-GB/16-GB/32-GB DDR4 SO-DIMM @ 3200 MHz
 - Networking
     - M.2 PCIe/CNVi WiFi/Bluetooth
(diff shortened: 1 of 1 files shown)


jacobgkau:
Interesting. This was originally added to the lemp9 page by Jeremy in [this commit](https://github.com/system76/tech-docs/commit/e8709cdb3b6527fe195ae928679baa391aeeedf2) with the link being added in [this commit](https://github.com/system76/tech-docs/commit/c99878547034e255ccbc3a1c9682fc5552269ea7). The upstream Clevo service manual has the part number for the 16GB module in four locations (which is why I left it in for this new version), but I didn't notice that it was the wrong size.

image

What's more confusing, though, is that I flipped over the motherboard of our lab unit and found a different 16GB part number, `K4AAG165WA-BCWE`. Samsung's [product page](https://www.samsung.com/semiconductor/dram/ddr4/K4AAG165WA-BCWE/) for that one states it's a "sample" rather than "mass production" part.

soldered-ram-small

The product pages also state (under `Org.`) that the 16GB parts are `1G x 16`, and the 8GB parts are `1G x 8`. Is it possible that Clevo is purchasing 16GB batches from Samsung and just using half of them per motherboard produced? I think if we're going to change this, we'd want to confirm with them in any case. The upstream service manuals can definitely be wrong, but right now both the service manual and our hardware do not suggest the 8GB part number is correct.

leviport:
Interesting, I think splitting batches is exactly what they're doing. I'm sure the chips themselves would otherwise be the same as the 8x1GB kits of the same speed. I've pushed a correction that I think explains it well enough. Let me know if you think that's sufficient.

jacobgkau (approved):
We know the memory speed increased to 3200MHz with the update to lemp10, and I see the `K4AAG165WA-BCWE` part available for sale on parts websites (in high quantities) despite Samsung's "sample" label, so I think it's safe to say that it's the part we're shipping and the Clevo schematics were not updated like they should have been. (This lab unit is already a later build with a production CPU instead of the original lab unit with a pre-production CPU, so I don't think it would be using pre-production RAM.)

PR #51, opened 2020-12-23
Add addw2

jacobgkau:
This PR adds the Adder WS 2 (addw2) to tech docs. This is the last current laptop that is not on tech docs.

<del>(Opening as draft because we still need to add AC adapter model numbers and barrel sizes to as many models as possible with this update, at support's request.)</del> Charger models & barrel sizes have been added.


src/models/addw1/README.md
```diff
@@ -25,6 +25,8 @@ The System76 Adder WS is a laptop with the following specifications:
     - Up to 64GB dual-channel DDR4 @ 2666 MHz, or
     - Up to 32GB dual-channel DDR4 @ 3000 MHz
 - Power
-    - 230W (19.5V, 9.23A) AC adapter
+    - 230W (19V, 9.23A) AC adapter
```
leviport, line 28: Are you sure the 230W charger is 9.23A? The 180W oryp6 charger says it's 9.23A so it seems like it should be higher. Serw12 has the same amperage listed, so that'll need updated as well if I'm right.

jacobgkau: Good catch, looks like that was wrong on addw1 and serw12 before, but it's fixed now. I'll try to keep in mind how electricity works when adding future chargers.

leviport (approved):

PR #52, opened 2020-12-31
Fix meer5 power adapter info

leviport:
I'm not sure I like the model disclaimer, so holler if you have a better way to word it and I'll be happy to change it.


thomas-zimmerman (approved):
That looks good. I can't come up with better language either.

PR #53, opened 2021-01-13
Pang10 initial

leviport:


leviport:
Don't merge until pang10 launches

src/models/pang10/README.md
```diff
@@ -0,0 +1,47 @@
...
+The System76 Pangolin is a laptop with the following specifications:
+
+- CPU
+    - Supports AMD Ryzen (3rd Gen) Mobile CPUs
+        - [Ryzen 5 4500U](https://www.amd.com/en/products/apu/amd-ryzen-5-4500u)
+        - [Ryzen 7 4700U](https://www.amd.com/en/products/apu/amd-ryzen-7-4700u)
+- Firmware
+    - System76 Firmware (non-open)
+- EC
+    - ITE IT8587
```
jacobgkau, line 18: Pages 1-8 (mainboard overview, top) and B-23 (schematic diagrams) of the upstream Clevo manual refer to the ITE IT5570. Is the 8587 what's actually in the machine?

leviport: You're right, it's the 5570. I meant to circle back to that and look it up but forgot about it.

jacobgkau:
I specified the version and generation of the USB-C port and added the generation of the USB-A port. I also re-ordered them to be consistent with the ordering on other products. (I think it's strange we're referring to the USB-A port as 3.1 Gen 1 when the Clevo manual says 3.2 Gen 1, but those are equivalent, so not an issue.)

For the USB-C port, we have not specified features that _aren't_ supported on previous products, but I'm not opposed to it (I'm sure it will be a common question.) I added charging to the list of non-present features.

leviport:
Nice, those are good changes. Thanks!

jacobgkau (dismissed):
There was a step in the keyboard removal that had its photo listed before the explanation. I moved the step before the photo, since that's how the rest of the page is written.

Also, the under-keyboard photo had the power button/LED connector labeled as yellow, but used #caf104, which looked more green to me (and the keyboard backlight connector was labeled green.) Several color websites confirmed that shade leans green: [1](https://www.color-hex.com/color/caf104), [2](https://encycolorpedia.com/caf104). I tried replacing it with a more ordinary yellow, but it got too difficult to see due to the brightness of that area in the photo; changing it to orange made it much easier to see.

That's all for my nitpicking. I'm approving now, but we can wait to merge until launch if you want.

jacobgkau:
@leviport Can we add the BIOS chip information to match the other laptop pages as added in https://github.com/system76/tech-docs/pull/62? It needs the flash chip model in the tech specs and, if possible, a highlight in the internal overview (the rest are highlighted black since we never used that color for anything else.)

jacobgkau (approved):
Just replaced the >20MB feature image w/ screen turned off with a compressed feature image w/ screen turned on, to better match the rest of the laptop feature images.

BIOS chip info looks good, this is ready to merge tomorrow.

PR #55, opened 2021-01-20
Update header to reflect validity for both r1 and r2

jacobgkau:
This section could use some work in general, but this change will tide over requests for r2 RAM configuration. We'll plan to start doing per-model entries like we have for massive-b1.2 and mega-r1.0 with the next revision of each version (AMD and Intel.)


leviport (approved):
Nice, thanks for jumping on this right away

PR #59, opened 2021-02-19
Add 15" gaze15 bottom panel info, and make corrections for keyboard removal

jacobgkau:
This PR makes the following changes:
- Adds a note at the top about connector/component locations potentially being different from what's pictured, due to differences between the 15" and 17" models.
- Adds a photo of the 15" bottom panel with screw locations and screw count (for both bottom panel removal and keyboard removal.)
- Corrects some info about the keyboard ribbons that was incorrectly carried forward from older Gazelle manuals (applies to both 15" and 17".)


storyaddict (approved):
Thanks for making these clarifications. These look good. :)

PR #62, opened 2021-03-02
Document BIOS flash chips

crawfxrd:
Need confirmation on location for most of the chips. I can't read the silkscreen on the pictures in the manuals or the pictures we have to know if it's the right chip.


src/models/oryp6/README.md
```diff
@@ -13,8 +13,8 @@ The System76 Oryx Pro is a laptop with the following specifications:
         - [Intel i7-10875H](https://ark.intel.com/content/www/us/en/ark/products/202329/intel-core-i7-10875h-processor-16m-cache-up-to-5-10-ghz.html)
 - Chipset
     - [Intel HM470](https://ark.intel.com/content/www/us/en/ark/products/203697/intel-hm470-chipset.html)
-- Firmware
-    - [System76 Open Firmware](https://github.com/system76/firmware-open)
+- BIOS
+    - MX25L12872F flash chip running [System76 Open Firmware](https://github.com/system76/firmware-open)
```
jacobgkau, line 17: The chip listed for oryp6 in the [firmware-open repo](https://github.com/system76/firmware-open/blob/master/models/oryp6/chip.txt) is `MX25L12805D`. I see that `MX25L12872F` is listed in the Clevo service manual. Should we always go with what's in the service manual, or did you find `MX25L12872F` somewhere else? The manual calls it `MX25L12872FM2I-10G`, how did you decide which character to stop at? (I'm trying to include this info in the oryp7 section and want to make sure I'm getting it right.)

crawfxrd: - MX25L12872F is the chip 
- MX25L12872FM2I-10G is the exact part number for the variant
  - see sections 18 and 19 in the Macronix datasheet for explanation
- MX25L12805D is the flashrom compatible name

jackpot51 (approved):

jacobgkau:
Confirmed locations & highlighted as many of these as I could with the machines and photos I currently have access to. serw12 & bonw14 cases are more enclosed and the chip's not visible, even in person. Corrected silkscreen label on gaze15 (seems to differ between different models.) Corrected chip model on galp4.

jacobgkau (approved):

PR #63, opened 2021-03-03
Add oryp7

jacobgkau:
This section adds the Oryx Pro 7 (oryp7) to tech docs.

It also adds:
- 2.5" drive instructions to the existing gaze15 section, as requested by support.
- Material information to the existing oryp6 section.


ahoneybun:
Are we sure it is a USB Type C with Thunderbolt 3? Every new model has been Thunderbolt 4. Having Thunderbolt 3 would make sense if it is a 10th Gen CPU though.

leviport:
10th gen Intel only support PCIe 3.0 @ahoneybun

src/models/oryp7/repairs.md
```diff
@@ -0,0 +1,200 @@
...
+**Tools required:** Cross-head (Phillips) screwdriver  
+**Time estimate:** 10 minutes  
+**Difficulty:** Easy <span style="color:green;">●</span>  
+
+### Steps to replace the M.2 drive:
+
+1. Follow the steps above to [remove the bottom cover](#removing-the-bottom-cover).
+2. Unscrew the retainer screw opposite the M.2 slot.
+
+![M.2 slots](./img/m2-slots.jpg)
```
ahoneybun, line 64: Since one port only works with M.2 NVMe drives can we highlight which port that is?

ahoneybun (changes requested):

src/models/oryp7/repairs.md
leviport: I mean, they're labeled on the motherboard.

jacobgkau: We state at the top of this section:

> The inner-most slot (closest to the battery) supports only PCIe NVMe Generation 3, and the outer-most slot (farthest away from the battery) supports either SATA III or PCIe NVMe Generation 3.

Do you not think this description is sufficient? I can see complaints about color-blindness accessibility by relying on colors for this kind of information where there's one item of either color. Also keep in mind that this information is physically silkscreened onto the motherboard, that this is the exact same wording the oryp6 manual is currently using, and that addw2, darp7, and serw12 also currently identify the slots by position instead of color.

ahoneybun: Sorry I'm blind @jacobgkau that should be fine. @leviport this is more about the photos having both slots filled so we can't see those labels.

leviport: My point was that when someone's installing a drive, they'll be able to read the labels on the motherboard. That seems like the only time knowing which is which would matter, right?

ahoneybun: That is correct. I think my point is that having the information is almost never bad so I just wanted to float that idea by you cool folks.

ahoneybun: Other then that everything else looks amazing as usual! <3

ahoneybun (approved):

ahoneybun:
With that recent change I think we are good to merge once QA approves right?

src/models/oryp7/repairs.md
jacobgkau: @ahoneybun I know we resolved this yesterday, but I later thought of adding labels directly on the image like this. I think it's clearer than the colors. What do you think?

src/models/oryp7/README.md
```diff
@@ -0,0 +1,58 @@
...
+
+The System76 Oryx Pro is a laptop with the following specifications:
+
+- CPU
+    - Supports Intel Comet Lake CPUs
+        - [Intel i7-10870H](https://ark.intel.com/content/www/us/en/ark/products/208018/intel-core-i7-10870h-processor-16m-cache-up-to-5-00-ghz.html)
+- Chipset
+    - [Intel HM470](https://ark.intel.com/content/www/us/en/ark/products/203697/intel-hm470-chipset.html)
+- BIOS
+    - MX25L12873F flash chip running [System76 Open Firmware](https://github.com/system76/firmware-open)
```
crawfxrd, line 17: Is this correct? U66 is shown as GD25R127D in the schematics. But MX25L12873F is listed below the GigaDevice chip and `chip.txt` has MX25L12805D...Does the chip package say Macronix?

jacobgkau: @crawfxrd I was trying to match what you provided for oryp6 (`MX25L12872F`) in https://github.com/system76/tech-docs/pull/62. The oryp6 Clevo manual shows:

oryp6-bios

The oryp7 Clevo manual shows:

oryp7-bios

I also found a [datasheet](https://media.digikey.com/pdf/Data%20Sheets/Macronix/MX25L12873F.pdf) for `MX25L12873F` similar to the one you linked for the `MX25L12872F`. I do see what you're talking about with the schematic being different.

> Does the chip package say Macronix?

No, but it does say `MX25L12873F`. Here is a picture:

P1030647

(This is labeled U66, one of the 6's is cut off but it's definitely the BIOS flash chip as I was able to flash it with a CH341A previously.)

bflanagin (approved):

PR #65, opened 2021-03-10
Add thelio-mira-r1 tech-docs

leviport:
But open a PR correctly this time


src/models/thelio-mira-r1.0/README.md
```diff
@@ -0,0 +1,81 @@
+# Thelio Mira (thelio-mira-r1.0)
+
+- [External Overview](./external-overview.md)
+- [Internal Overview](./internal-overview.md)
+- [Parts & Repairs](./repairs.md)
+
+![Thelio Mira](./img/mira-front.jpg)
```
ahoneybun, line 7: Can we use this instead?

https://assets.system76.com/products/thelio-mira-r1/feature.png?fit=cover&width=500

src/models/thelio-mira-r1.0/external-overview.md
```diff
@@ -0,0 +1,25 @@
...
+  - Line In / Side Speaker Out
+
+### Power Switch Guide:
+
+- 1: On
+- 0: Off
+
+**Note:** The BIOS Flash button (next to the CMOS reset button) is not needed when performing UEFI firmware updates via the System76 Firmware Manager. Use of the BIOS Flash button is not recommended.
+
+**Note:** The HDMI and DisplayPort will only work if your CPU supports integrated graphics, like the AMD 3200g and 3400g. If your CPU does not support integrated graphics, use the display outputs on your graphics card instead.
```
ahoneybun, line 25: We may want to change 3200g to 3200G and 3400g to 3400G.

ahoneybun (changes requested):
I just had some very minor comments for some changes.

src/models/thelio-mira-r1.0/README.md
```diff
@@ -0,0 +1,81 @@
...
+        - [Ryzen 7 5800X](https://www.amd.com/en/products/cpu/amd-ryzen-7-5800x)
+        - [Ryzen 9 5900X](https://www.amd.com/en/products/cpu/amd-ryzen-9-5900x)
+        - [Ryzen 9 5950X](https://www.amd.com/en/products/cpu/amd-ryzen-9-5950x)
+- Motherboard
+    - [ASUS ROG Strix X570-E Gaming](https://rog.asus.com/us/motherboards/rog-strix/rog-strix-x570-e-gaming-model/) running System76 Firmware (non-open)
+    - AMD X570 chipset
+- Daughterboard
+    - [Thelio Io](https://github.com/system76/thelio-io) board running [open-source firmware](https://github.com/system76/thelio-io-firmware)
+- GPU options
+    - One or two GPUs, depending on configuration and power requirements (3x PCIe 3.0/4.0 x16 slots)
```
jacobgkau, line 25: Is there a reason we're listing the slots as `3.0/4.0`? The 4.0 spec is backwards-compatible with 3.0, so we should only have to list 4.0, correct? (Mega just calls them 4.0.)

src/models/thelio-mira-r1.0/internal-overview.md
```diff
@@ -0,0 +1,26 @@
+# Thelio Mira (Internal Overview)
+
+## Side overview:
+
+![Internal overview (side)](./img/internal-labeled.png)
+
+- CPU cooler is highlighted in blue
+- RAM slots are highlighted in red
+- PCIe slots are highlighted in green
```
jacobgkau, line 9: Is there a reason we're not highlighting/listing the x1 slots? GPUs would typically only use the x16 slots, but the two x1 slots should be usable depending on the GPU configuration, right?

src/models/thelio-mira-r1.0/README.md
leviport: Only because I've seen questions about backwards compatibility before. Just saying 4.0 should be enough though, so we can change it if you think it looks better.

jacobgkau: Looks like it's to distinguish that some of the slots will run at 3.0 with older-generation processors? The motherboard manual only lists the top two as 3.0/4.0, and lists the bottom as 4.0-only (indicating it will run at 4.0 even with 1st gen processors.) See pages 1-6 and 1-7. https://dlcdnets.asus.com/pub/ASUS/mb/SocketAM4/ROG_STRIX_X570-E_GAMING/E15826_ROG_STRIX_X570-E_GAMING_UM_v2_WEB.pdf

src/models/thelio-mira-r1.0/internal-overview.md
leviport: I just didn't think it was important, but I can add that. I still have my original SVG, so I'll make the modifications and push a new image.

src/models/thelio-mira-r1.0/README.md
```diff
@@ -0,0 +1,81 @@
...
+    - Up to 128GB (4x32GB) quad-channel ECC Unbuffered DDR4 DIMMs @ 3200 MHz
+    - Tested with the following RAM modules (may ship with other tested modules):
+        - [Kingston HX432C16FB3/8](https://www.kingston.com/dataSheets/HX432C16FB3_8.pdf)
+        - [Kingston HX432C16FB4/64](https://www.kingston.com/dataSheets/HX432C16FB3K4_64.pdf)
+        - [Kingston HX432C16FB3K4/128](https://www.kingston.com/dataSheets/HX432C16FB3K4_128.pdf)
+        - [Micron MTA9ASF2G72AZ-3G2B1](https://www.micron.com/products/dram-modules/udimm/part-catalog/mta9asf2g72az-3g2/mta9asf2g72az-3g2b1)
+        - [Micron MTA18ASF4G72AZ-3G2B1](https://www.micron.com/products/dram-modules/udimm/part-catalog/mta18asf4g72az-3g2/mta18asf4g72az-3g2b1)
+- Networking
+    - 1x 2.5-Gigabit Ethernet port (Realtek RTL8125-CG 2.5G LAN)
+    - 1x 1-Gigabit Ethernet port (Intel I211-AT, 1 x Gigabit LAN Controller)
```
jacobgkau, line 59: I'm a little confused by how this is listed. I see that the motherboard manual lists:
image
What does `1 x Gigabit LAN Controller` mean on this line? It looks like the Realtek port uses its own controller, any objections to dropping this redundant detail so it looks like this instead?
```
- 1x 2.5-Gigabit Ethernet port (Realtek RTL8125-CG)
- 1x 1-Gigabit Ethernet port (Intel I211-AT)
```
Or am I reading this wrong, and you're trying to say that both ports use a single Intel controller?

leviport: Yeah that's totally weird. I have no idea why I wrote it. We should definitely change it to the way you wrote it.

src/models/thelio-mira-r1.0/README.md
```diff
@@ -0,0 +1,81 @@
...
+    - Three power supply options:
+        - [EVGA 220-B5-0650 (650W)](https://www.evga.com/products/product.aspx?pn=220-B5-0650-V1)
+        - [EVGA 220-G5-0750 (750W)](https://www.evga.com/products/product.aspx?pn=220-G5-0750-X1)
+        - [EVGA 220-G5-1000 (1000W)](https://www.evga.com/products/product.aspx?pn=220-G5-1000-X1)
+    - C13 power cable
+- Sound
+    - 3.5mm line out, line in, microphone jacks
+    - Optical S/PDIF out
+    - Up to 7.1-channel audio output
+    - AMD Starship/Matisse HD Audio Controller
```
jacobgkau, line 71: The motherboard manual lists ROG SupremeFX S1220A for the codec/chipset. We have audio codecs/chipsets listed for Massive and Mega, I think we should try to carry that level of detail forward.

I'm seeing that the S1220A may be a Realtek ALC1150. Do you have any lspci/lsusb or alsainfo output from the Mira to confirm the best way to list this?

leviport: I didn't save the output, and I just returned Mira to the office to be rebuilt with the new chassis revision, so I can't confirm that. I got the "Starship/Matisse" line from lspci, but I might have grabbed the wrong bit of info. Feel free to update if you think it needs it.

leviport:
I added a white background on the rear port diagram picture in the external overview. The dark text labels become very hard to read when switching to the dark themes. We should probably update that on the other pages at some point, but we can get to that later.

src/models/thelio-mira-r1.0/README.md
leviport: Looks to me like the manual lists all three x16 slots as 3.0/4.0 and the x1 slots as 4.0 (which _should_ still be backwards compatible because 4.0 only x1 slots makes no sense). I'm leaning towards changing it to `3x PCIe 4.0 x16 slots` for the sake of brevity.

jacobgkau:
This looks okay to me at this point. @leviport Are we planning to launch this with outdated photos? I'd of course prefer we have accurate photos (with the correct number of PCIe cutouts), but since we don't really show the bottom PCIe slot in action anyway, I wouldn't be opposed to launching now if we're not going to get an accurate lab unit very soon.

leviport:
Yeah, I was planning to launch with the outdated photos and swapping them out when I got the updated chassis. We're currently estimating it being ready later this week.

I'll add a note on the external overview.

jacobgkau (approved):

PR #75, opened 2021-05-14
Switch the AC power adapter model numbers, the description was...

thomas-zimmerman:
correct but the models where switched.


leviport (approved):

leviport:
Thanks for the correction!

PR #76, opened 2021-05-20
removed the phillip driver needed for external battery removal

ahoneybun:
A customer pointed out that the external battery is not needed for removal of the external battery so this is to update that.


jacobgkau (approved):
Please add the spaces back to the end of the line, the line break is currently missing in the rendered version:

image

jacobgkau:
Fixed it myself since I accidentally approved.

jacobgkau (approved):

PR #77, opened 2021-05-26
Add meer6

jacobgkau:
The service manual is complete and tech specs should be correct. Waiting on an M.2 2242 SATA drive to test with so we know how to list that (e.g. whether it works at the same time as the 2280 NVMe slot), and feature/port photos.


ahoneybun (approved):
Everything looks great as usual! :tada:

PR #80, opened 2021-06-04
Add meer5 graphics info

jacobgkau:
This adds a graphics section to the meer5 tech specs.

With oryp7, I called this section "Graphics" instead of "GPU" to enable better organization of the GPU options, built-in display options, and video output ports as sub-categories. The meer5 section follows this new formatting.


leviport (approved):
:metal:

PR #81, opened 2021-06-10
Launch initial

leviport:


src/models/launch_1/customization.md
```diff
@@ -0,0 +1,42 @@
+# Launch (Usage & Customization)
+
+The Launch Configurable Keyboard can be customized and personalized in a variety of ways. It is recommended to remove unplug all USB cables and devices from your Launch before changing keycaps or switches.
+
+- [Connecting and Using the Launch](#connecting-and-using-launch)
```
jacobgkau, line 5: The section title does not have `the`, they should probably match unless there's a specific reason for them not to.

(The only place I've done this in the past is `Replacing the WiFi/Bluetooth module` linking to `Replacing the wireless card`, but the reason for that was so if someone's looking for WiFI/Bluetooth, they'll see it easier.)

src/models/launch_1/customization.md
```diff
@@ -0,0 +1,42 @@
...
+- [Removing and Installing Switches](#removing-and-installing-switches)
+- [Removing the PCB](#removing-the-pcb)
+
+## Connecting and Using Launch
+
+![Launch ports](./img/launch-ports.png)
+
+Using either the USB-C to USB-C cable or the USB-A to USB-C cable, connect to your computer via your Launch's center USB type C port. Due to the directionality of USB, your Launch will only work when connected to your computer via the center port. Launch should now be ready to type on, and the USB ports on the back are ready to connect devices to.
+
+## Using the System76 Keyboard Configurator
```
jacobgkau, line 18: This is very much a software thing. Does this info belong here instead of in readme files within the configurator repository or a support article? (Not saying it doesn't, just bringing it up. We omitted the BIOS screenshots from tech docs for a reason.)

A potential reason it doesn't fit here is that the configurator, being software, will inevitably see updates, at which point this section would have to be updated; and if the configurator is used on future keyboard revisions, the software updates would still apply to older hardware revisions, so it would need to be updated in every section every time. Also, the configurator applies to Open Firmware laptops, so would we add this section to every Open Firmware laptop manual too?

leviport: Makes sense. I've pulled out the configurator sections and will make a PR in support docs for those instead.

src/models/launch_1/README.md
```diff
@@ -0,0 +1,40 @@
...
+  - PBT plastic
+  - Dye sublimated legend
+  - XDA profile
+- Layout
+  - Modified ANSI US QWERTY
+- Custom Configurations
+  - Customize layout and lighting in firmware with the [System76 Keyboard Configurator](https://github.com/pop-os/keyboard-configurator)
+  - [Available on Linux, Windows, and macOS](https://system76.com/accessories/launch/download)
+- Integrated Hub
+  - 2x USB 3.2 Gen 2 Type-C (Up to 10 Gbps)
```
jacobgkau, line 30: We don't provide "Up To" for the USB ports on any other products; the port generation covers this information. Is it okay to remove that?

leviport: Yeah I think that's fine

src/models/launch_1/customization.md
```diff
@@ -0,0 +1,95 @@
...
+The Launch Configurable Keyboard can be customized and personalized in a variety of ways. It is recommended to unplug all USB cables and devices from your Launch before changing keycaps or switches.
+
+- [Connecting and Using Launch](#connecting-and-using-launch)
+- [Removing and Installing Keycaps](#removing-and-installing-keycaps)
+- [Removing and Installing Switches](#removing-and-installing-switches)
+- [Removing the PCB](#removing-the-pcb)
+
+## Connecting and Using Launch
+
+![Launch ports](./img/launch-ports.png)
```
jacobgkau, line 12: This image highlights a number of downstream ports that would not work for connecting to the computer. Can we get a photo with only the port we're talking about highlighted? (I would take a head-on photo and highlight with a box in Inkscape, but if we really want to reuse the port diagram photo, just removing the rest of the labels would also work.)

jacobgkau: You could also highlight them all with boxes, but make the center one a different color. That way, you can more clearly point out which one to use for the computer and which ones can be used for downstream devices.

src/models/launch_1/customization.md
```diff
@@ -0,0 +1,95 @@
...
+
+**Tools required:** Keycap puller, MX-style switch puller, and cross-head (Phillips) screwdriver  
+**Time estimate:** 50 minutes  
+**Difficulty:** High <span style="color:red;">●</span>  
+
+If the PCB in your Launch ever becomes damaged, it can be replaced using these instructions. It is also recommended to completely remove the PCB if you are changing all of the switches on your Launch at the same time. This makes switch installation much easier and makes it much less likely to bend switch pins or damage the PCB.
+
+Begin by [removing all of the keycaps](#removing-and-installing-keycaps), and then [remove all of the switches](#removing-and-installing-switches). Next, turn the keyboard over and remove the eight screws that hold on the bottom panel and remove the bottom panel. Lifting the bottom panel out of the chassis may be easier with the magnetic lift bar installed.
+
+![Removing bottom panel screws](./img/bottom-panel-screws.jpg)
```
jacobgkau, line 76: We don't typically show the usage of a screwdriver on computer service manuals. It makes sense for the keycap picker because that's an unusual tool, but is it necessary here? A more head-on photo without having to hold the screwdriver there might make it easier to get good lighting on the screws.

jacobgkau:
@leviport I broke down the steps on the repairs page so they're more step-by-step. Could you take a quick look through to cross-check my changes, since I'm a bit of a keyboard noob and haven't really touched a Launch? If you're okay with those changes, then this should be good to go.

leviport:
Those changes look great, thanks!

PR #86, opened 2021-07-07
Add LCD panel model numbers

### LINKED ISSUE #6: lemp9: LCD panel information
Having basic info about the LCD panel, such as size, resolution, etc. seems helpful. Going even further and including the make and model of the panel would be even better.

At some point, having instructions for replacing the LCD panel would be fantastic. Shipping machines back for repairs isn't always an option for all users, and many are handy enough to perform such a repair themselves.

jacobgkau:
Closes #6.

This adds at least one compatible LCD panel model number for each laptop (I listed all numbers we're aware of, so some models have more than one option listed.) Every model number is accompanied by `or equivalent`, so we should be safe in case our supply chain causes any of these to change, which does happen from time to time.

This also reorganizes all `Graphics` sections to use the newer format, and makes the following small corrections:
- lemp9 and lemp10 were both 14.0", not 14.1".
- serw12 was/is 144Hz (we originally marketed 120Hz, but both panels say 144Hz.)
- Added 1650 Ti to galp5.

### DIFF
--- a/src/models/addw1/README.md
+++ b/src/models/addw1/README.md
@@ -22,9 +22,9 @@ The System76 Adder WS is a laptop with the following specifications:
     - eDP display: 15.6" 3840x2160@60Hz OLED
         - LCD panel: Samsung ATNA56WR06 (or equivalent)
     - External video output:
-        - 1 x HDMI (w/HDCP)
-        - 1 x Mini DisplayPort 1.3
-        - 1 x DisplayPort 1.3 over USB 3.1 Type C
+        - 1x HDMI (w/HDCP)
+        - 1x Mini DisplayPort 1.3
+        - 1x DisplayPort 1.3 over USB 3.1 Type C
 - Memory
     - Up to 64GB dual-channel DDR4 @ 2666 MHz, or
     - Up to 32GB dual-channel DDR4 @ 3000 MHz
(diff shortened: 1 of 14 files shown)


src/models/addw1/README.md
```diff
@@ -17,12 +17,14 @@ The System76 Adder WS is a laptop with the following specifications:
...
-- GPU
-    - NVIDIA GeForce RTX 2070
-    - eDP 15.6-inch 3840x2160 OLED
-    - 1 x HDMI (w/HDCP) output
-    - 1 x Mini DisplayPort (1.3) output
-    - 1 x DisplayPort (1.3) over USB 3.1 Type C output
+- Graphics
+    - GPU: NVIDIA GeForce RTX 2070
+    - eDP display: 15.6" 3840x2160@60Hz OLED
+        - LCD panel: Samsung ATNA56WR06 (or equivalent)
```
leviport, line 23: Should "LCD" be changed to "OLED"?

(also applies to addw2 and the OLED option on 15" oryp7)

leviport:
Other than the LCD/OLED uncertainty, I think this looks great!

src/models/addw1/README.md
jacobgkau: I guess that makes a lot of sense, done with 6f15fe8. (I'm now questioning whether we need LCD/OLED on both lines instead of just one, but leaving it out from either line doesn't seem right, so I'll leave it for now.)

leviport (approved):
:metal:

PR #87, opened 2021-07-14
Add gaze16

jacobgkau:
Adds the Gazelle 16 to tech docs. (Also adds a missing header in the gaze15 repairs section.)


src/models/gaze16/external-overview.md
```diff
@@ -0,0 +1,122 @@
...
+|![Storage LED](./img/led-storage.png)   |Green          |Storage drive activity           |
+
+## Keyboard shortcuts:
+
+The Gazelle has the following actions available using the Fn and Function keys:
+
+|Key                        |Shortcut|Action                                        |
+|---------------------------|--------|----------------------------------------------|
+|![Fn-F1](./img/fn-f1.png)  |Fn+F1   |Toggle trackpad                               |
+|![Fn-F2](./img/fn-f2.png)  |Fn+F2   |Toggle built-in LCD<br/>(non-OLED models only)|
```
ahoneybun, line 108: Since there isn't an OLED model do we need this or just future proofing the tech-docs?

src/models/gaze16/internal-overview.md
```diff
@@ -0,0 +1,25 @@
...
+## Bottom components and connectors:
+
+![Internal Components](./img/components-highlighted.jpg)
+
+- Webcam/microphone connector is highlighted in pink
+- Fan connectors are highlighted in orange
+- LCD panel connector is highlighted in red
+- Wireless card is highlighted in blue
+- RAM is highlighted in dark green
+- M.2 SSDs are highlighted in yellow
```
ahoneybun, line 12: Since the slots are like this:
1x M.2 (PCIe NVMe Gen 4)
1x M.2 (PCIe NVMe Gen 3 or SATA)

Should we highlight this differently then?

ahoneybun: I see that we do that in the Parts & Repair section so maybe copy that here as well?

ahoneybun (changes requested):
LGTM other then some minor changes!

src/models/gaze16/external-overview.md
jacobgkau: This was left over from the oryp7 section I copied as a template, thank you for catching it. Removed with 78fd1e2.

ahoneybun (approved):
LGTM with those changes!

PR #90, opened 2021-08-12
Add pang11

leviport:


leviport:
I'm still unsure about the LCD panel info. Currently trying to verify whether it's the same as pang10.

src/models/pang11/README.md
```diff
@@ -0,0 +1,51 @@
...
+- [External Overview](./external-overview.md)
+- [Internal Overview](./internal-overview.md)
+- [Parts & Repairs](./repairs.md)
+
+![Pangolin](./img/pang11.webp)
+
+The System76 Pangolin is a laptop with the following specifications:
+
+- CPU
+    - Supports AMD Ryzen (3rd Gen) Mobile CPUs
```
jacobgkau, line 12: We listed 3rd Gen for both the 4xxx from pang10 and the 5xxx on pang11. Shouldn't they be 4th and 5th gen, or should we refer to them some other way (codenames)?

leviport: The 5500U and 5700U are still Zen2, which is gen 3. It might not be the most useful piece of info, but it was present on pang10 so I just left it.

jacobgkau: I'm seeing generational names of Lucienne for 5xxx and Renoir for 4xxx, both based on the Zen 2 architecture. Would those names be more useful than the number, do you think? (I can barely find anyone referring to generation numbers for AMD, and of those I can find, I've seen a few retailers incorrectly referring to "5th gen".)

leviport: Yeah that does sound a lot better. I'll make the change.

jacobgkau (approved):
This looks ready to merge once pang11 is shipping.

PR #95, opened 2021-08-28
Correct duplicate i7 in power spec

jakern:
Said 'i7 and i7' but should be 'i5 and i7'


jacobgkau (approved):
Thank you for the correction!

PR #97, opened 2021-09-13
update tools required for fan replacement

sed-i:
add tweezers, rubbing alcohol to tools required for fan replacement

### DIFF
--- a/src/models/oryp7/repairs.md
+++ b/src/models/oryp7/repairs.md
@@ -138,9 +138,9 @@ If the CPU/GPU fans become noisy and cleaning them out doesn't fix the issue, yo
 
 Depending on your climate and the age of the machine, it may be necessary to apply new thermal paste between the CPU/GPU and the heatsink. Thermal paste helps facilitate effective heat transfer between the CPU/GPU and the cooling equipment. These instructions can also be used in the unlikely event your heatsink needs replacing.
 
-**Tools required:** Cross-head (Phillips) screwdriver  
-**Time estimate:** 20 minutes  
-**Difficulty:** High <span style="color:red;">●</span>  
+**Tools required:** Cross-head (Phillips) precision screwdriver, tweezers, rubbing alcohol
+**Time estimate:** 20 minutes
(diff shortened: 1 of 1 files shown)


jacobgkau (changes requested):
Neither of these things are required for replacement. Thermal paste can be wiped off with a dry paper towel if necessary, and tweezers are not needed at all.

sed-i:
@jacobgkau,
1. Rubbing alcohol (IPA) is already mentioned in the body text itself. BTW, without IPA, wiping the thermal paste takes a considerable time and inexperienced users may damage something while they're at it.
2. One of the black fan screws is very difficult to position without tweezers. A magnetic screwdriver might do the trick too, but tweezers are very handy for placing/fetching screws and also for manipulating the paper towel for wiping thermal paste.

Feel free to close this if you disagree.

jacobgkau:
> 1. Rubbing alcohol (IPA) is already mentioned in the body text itself. BTW, without IPA, wiping the thermal paste takes a considerable time and inexperienced users may damage something while they're at it.

I would be more concerned about a user damaging their system with liquid rubbing alcohol than with a paper towel. As you said, it is mentioned as an optional suggestion within the body text, but it is by no means a requirement. (I have taken the thermal paste replacement photos for most of the models currently in tech-docs, and I have never used rubbing alcohol.)

> 2. One of the black fan screws is very difficult to position without tweezers. A magnetic screwdriver might do the trick too, but tweezers are very handy for placing/fetching screws and also for manipulating the paper towel for wiping thermal paste.

I can't say that I've ever run into this. Once again, I've taken apart almost all current models while creating the documentation (including the oryp7 being referred to here), and I've been able to screw in all of the screws using only a screwdriver.

I'm going to close this because neither of these are hard requirements. A screw organizer, a plastic spudger to help remove the bottom panel, a magnifying glass to easier read labels on the motherboard, and multiple sizes of "precision" screwdriver are all things that may aid in computer repair, but they are not things that a user would need to purchase in order to service this model. Thank you for the suggestions, though.

PR #101, opened 2021-09-22
Add thelio-mira-b1.0

jacobgkau:
This PR adds the Thelio Mira b1.0 to tech docs, including tech specs, external overview, internal overview, and service manual.

In the service manual, I tired a new format for the Thelio-IO wiring information. Rather than taking a solely diagnostic approach, it attempts to explain how to wire from scratch, which should hopefully be more clear for both troubleshooting and general rebuilding.


src/models/thelio-mira-b1.0/README.md
```diff
@@ -0,0 +1,89 @@
...
+        - [NVIDIA Quadro RTX 6000](https://www.nvidia.com/en-us/design-visualization/quadro/rtx-6000/) (maximum 2)
+            - 4x DisplayPort 1.4, DisplayPort over USB-C (VirtualLink)
+        - [NVIDIA Quadro RTX 8000](https://www.nvidia.com/en-us/design-visualization/quadro/rtx-8000/) (maximum 2)
+            - 4x DisplayPort 1.4, DisplayPort over USB-C (VirtualLink)
+- Expansion
+    - 2x PCIe 4.0/3.0 x16 (GPU slots)
+    - 1x PCIe 4.0/3.0 x16 (additional slot; single-height cards only)
+    - 1x PCIe 3.0 x4
+- Memory
+    - Up to 128GB (4x32GB) quad-channel ECC Unbuffered DDR4 DIMMs @ 3200 MHz
```
leviport, line 54: I believe these CPUs only support dual-channel non-ECC RAM

src/models/thelio-mira-b1.0/README.md
```diff
@@ -0,0 +1,89 @@
...
+    - Wi-Fi 6 (Intel AX200)
+- Power
+    - C19 power cable
+    - 650W minimum PSU
+        - Dual-GPU configurations require 750W or 1000W, depending on GPU power requirements
+    - Tested with the following PSU models (may ship with other tested models):
+        - [EVGA 650W B5 80+ Bronze](https://www.evga.com/Products/Specs/PSU.aspx?pn=9C00E82F-02D0-4706-A361-78D42787345B)
+        - [EVGA SuperNOVA 750W G5 80+ Gold](https://www.evga.com/Products/Specs/PSU.aspx?pn=A7699F25-A614-4D88-A339-1DDF3D24615E)
+        - [EVGA SuperNOVA 1000W G5 80+ Gold](https://www.evga.com/products/Specs/PSU.aspx?pn=71d331b7-9306-4218-987f-875fd422fd44)
+        - [EVGA SuperNOVA 1600W T2 80+ Titanium](https://www.evga.com/Products/Specs/PSU.aspx?pn=50804BF2-4D32-4E90-B338-C81CA63AF4E2)
```
leviport, line 70: Were there configs that shipped with the 1600W PSU? I thought we only went up to 1000W on either mira.

src/models/thelio-mira-b1.0/README.md
```diff
@@ -0,0 +1,89 @@
...
+- Sound
+    - 3.5mm line out, line in, microphone jacks
+    - Up to 7.1-channel audio output
+    - Realtek ALC4080 audio chipset (with Savitech SV3H712 amplifier)
+    - HDMI, DisplayPort, USB-C DisplayPort audio (depending on GPU)
+- Storage
+    - 1x M.2 (PCIe NVMe Gen 4), M key
+        - Backwards compatible with PCIe NVMe Gen 3.
+        - Only active with 11th Gen CPUs.
```
leviport, line 79: I could see this being interpreted as "slot is only active with 11th gen CPUs" instead of "Gen 4 is only active with 11th gen CPUs". Maybe it should be more specific?

jacobgkau: The slot _is_ only active with 11th Gen CPUs, according to the motherboard manual.
image

jacobgkau: I originally had the active line above the compatibility line, but switched them because the compatibility related more to the line above than whether it's active or not. I can switch them back if that would clarify this further.

src/models/thelio-mira-b1.0/README.md
jacobgkau: This was definitely wrong, fixed.

src/models/thelio-mira-b1.0/README.md
leviport: Oh ok, I thought it just fell back to 3.0 with a 10th gen CPU. No need to change it, I was just mistaken on the actual functionality of the slot.

src/models/thelio-mira-b1.0/internal-overview.md
```diff
@@ -0,0 +1,33 @@
...
+## Side overview:
+
+![Internal overview (side)](./img/internal-overview.webp)
+
+- CPU cooler & CPU cooler fan are highlighted in solid green
+- CPU fan connectors are highlighted in solid pink
+- RAM slots are highlighted in dashed red
+- 2.5" SATA drive slot cover is highlighted in dashed pink
+- M.2 slots are highlighted in dashed cyan (located underneath the heatsinks)
+- PCIe slots are highlighted in yellow
```
leviport, line 12: The PCIe 3.0 x4 slot is highlighted in Purple, should that be added to the list? 

I'm also not sure the top PCIe slot should be highlighted with dashed yellow if we aren't going to mention what dashed yellow means anywhere. Dashed vs solid lines of the same color refer to unrelated components in some cases, so I feel like it should be mentioned if it's going to be a different line style.

jacobgkau: > The PCIe 3.0 x4 slot is highlighted in Purple, should that be added to the list?

Yes, added that to the list and relabeled the existing entry.

> I'm also not sure the top PCIe slot should be highlighted with dashed yellow if we aren't going to mention what dashed yellow means anywhere. Dashed vs solid lines of the same color refer to unrelated components in some cases, so I feel like it should be mentioned if it's going to be a different line style.

The reason for dashes in that case was to signify that it's behind the GPU; it otherwise looks like we're drawing a box around nothing since the other two boxes have the slots visible. (For the Mega and Massive internal overviews, I removed the GPUs as well as the 2.5" drive covers for this photo, unlike the mira-r1 section, and I'm thinking it should probably go back to that style for future sections.)

I could just add "Top slot (dashed yellow) is behind the GPU", could change it back to solid yellow and move it up so it's at the top of the GPU (overlapping the top M.2 slot's box), or I could retake the photo without the GPU installed, depending on how severe we think this is.

I will point out that in any other case where dashed vs. solid matters, we don't just label the dashed color, we also refer to the solid color using the word "solid." This line says only "yellow" because it includes all stroke styles of yellow.

leviport: >I could just add "Top slot (dashed yellow) is behind the GPU"
I like that the best. Fair point regarding "solid", but I do like it being a bit more explicit so nobody's left wondering.

src/models/thelio-mira-b1.0/repairs.md
```diff
@@ -0,0 +1,424 @@
...
+2. Unscrew the two screws securing the drive bay's cover.
+
+![SATA slot cover](./img/sata-slot-cover.webp)
+
+3. If you are adding a new drive, pop out the black plastic ring on the top crossbar and slide out four screws (per drive).
+
+![Storage drive screws](./img/storage-drive-screws.webp)
+
+4. Insert four screws into each 2.5" storage drive you wish to install.
+5. Slide each 2.5" drive into one of the slots leading to the Thelio-IO boards.
```
leviport, line 56: Should "boards" be singular?

leviport: I changed it in my last commit. We can change it back if it should be plural.

leviport:
Looks great! After the PCIe slot labeling tweak, have a look at the last couple commits I made. If they look good to you, I'll approve and we can merge it.

src/models/thelio-mira-b1.0/internal-overview.md
jacobgkau: Done.

src/models/thelio-mira-b1.0/repairs.md
```diff
@@ -151,11 +151,11 @@ Thelio Mira supports up to two dual-slot GPUs. All PCIe x16 slots run at Gen 4 s
 
 ![GPU brace screws](./img/gpu-brace-plastic-screws.webp)
 
-_The plastic GPU braces' primary function is to prevent damage during shipping. The system can be run without the braces if they don't fit an upgraded card; the back PCIe bracket provides primary support for the GPUs._
+_The GPU brace assembly's primary function is to prevent damage during shipping. The system can be run without the brace if it doesn't fit an upgraded card; the back PCIe bracket provides primary support for the GPUs._
```
jacobgkau, line 154: This line's revision is more than plural -> singular, this is changing from talking about the plastic pieces to talking about the entire metal brace assembly. Would we not still recommend putting the metal brace in to support the CPU shroud even if the plastic pieces are removed to make way for a larger GPU?

leviport: I don't think the CPU shroud needs additional support in situations where the GPU doesn't need additional support. It's a lot more sturdy and a lot less delicate. We can change it back if you think it's better to keep it installed though.

jacobgkau: @leviport I've revised this section. I suppose people are free to recognize for themselves that they can place a partial GPU brace back into their system if they want to. The more important point of the note is to assure customers that not having a fitting GPU brace is not an issue for normal operation of the system.

Also added a highlight for the depth adjustment screws/screwholes. (I thought about whether the rest of the screwholes for the height adjustment should also be highlighted to match, but I think it looks clear as-is.)

jacobgkau: Fixed the typo you pointed out, thank you.

leviport (approved):

PR #104, opened 2021-10-06
Add oryp8

jacobgkau:
This PR adds the Oryx Pro 8 (oryp8) to tech docs.

Significant changes from oryp7:
- Keyboard now uses an adhesive strip instead of screws
- SD card reader has been removed
- CPU is now 11th gen/Thunderbolt port is now version 4
- BIOS chip model number has changed

For the first time, we'll now have two generations of the same product in the "Previous Models" section of tech docs. For now, I'm ordering the generation numbers in descending order (most recent first.) Even though that's opposite from alphabetical order, I'm assuming the likelihood someone is looking for their version will go down the older it gets.


src/models/oryp8/README.md
```diff
@@ -0,0 +1,62 @@
+# Oryx Pro (oryp8)
+
+- [External Overview](./external-overview.md)
+- [Internal Overview](./internal-overview.md)
+- [Parts & Repairs](./repairs.md)
+
+![Oryx Pro](./img/oryp8.webp)
```
leviport, line 7: Looks like this image still has the SD card reader. That should probably be updated

PR #110, opened 2021-12-30
add missing links

ahoneybun:
This PR starts the process of getting the Thelio Major tech docs to match the other Thelio tech docs.


jacobgkau (approved):
These sections are still a long ways off from the rest (most notably, B2 and R2 should not be listed in the same section since they are different models), but the link addition is fine.

PR #112, opened 2022-02-01
Add networking specs for meer5

ahoneybun:
This PR does the following:
- Adds link for Ethernet controller
- Adds link for Wi-Fi card


jacobgkau (approved):
This is also adding the Bluetooth version, which matches the meer6 section's formatting.

PR #116, opened 2022-03-09
lemp9: Add version number for DisplayPort over USB-C

jacobgkau:
The lemp9 has DisplayPort 1.2 over USB-C, according to the upstream hardware manufacturer's manual.

CC: @ahoneybun


ahoneybun (approved):
Changes look good live. 

Screenshot from 2022-03-09 10-51-38

Issue #122, opened 2022-05-01
Confusing to separate Previous from Current models


ctrlcctrlv:

jacobgkau:
Is there any elaboration on this issue? Current models are separated out so people with brand new machines aren't having to look through an entire list of historical models to find documentation. I'm not sure there's anything to be done to make this less confusing.

crawfxrd:
This would be a good start:

```diff
diff --git a/book.toml b/book.toml
index 30ef0ed5e038..9e9724c515b8 100644
--- a/book.toml
+++ b/book.toml
@@ -14,3 +14,7 @@ theme = "theme"
 additional-css = ["theme/variables.css"]
 git-repository-url = "https://github.com/system76/tech-docs"
 git-repository-icon = "fa-github"
+
+[output.html.fold]
+enable = true
+level = 1
```

PR #124, opened 2022-05-04
Make sidebar easier to read

crawfxrd:
Collapse list so only the models are shown and remove section labels.

Resolves: #122

### DIFF
--- a/book.toml
+++ b/book.toml
@@ -14,3 +14,8 @@ theme = "theme"
 additional-css = ["theme/variables.css"]
 git-repository-url = "https://github.com/system76/tech-docs"
 git-repository-icon = "fa-github"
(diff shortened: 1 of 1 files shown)


jacobgkau (approved):
I don't see how this affects #122 very much, but it looks cleaner. Before vs. after:

image

The arrow can be used to open a section manually, or the section opens automatically when navigating into the section (and closes when navigating out of that section).

crawfxrd:
Considering they never explained what they meant, my interpretation was the amount of visual clutter made it hard to tell where each model was.

jackpot51:
We should really remove Components.

ctrlcctrlv:
Thank you, this is what I meant.

PR #130, opened 2022-05-23
Make power cord specs more consistent

leviport:
Also fix power cord type on addw1, addw2, and 2070 serw12 (C5 -> C13)

Also change "cable" to "cord" on desktops

### DIFF
--- a/src/models/addw1/README.md
+++ b/src/models/addw1/README.md
@@ -29,8 +29,8 @@ The System76 Adder WS is a laptop with the following specifications:
     - Up to 64GB dual-channel DDR4 @ 2666 MHz, or
     - Up to 32GB dual-channel DDR4 @ 3000 MHz
 - Power
-    - 230W (19.5V, 11.8A) AC adapter
-        - Included AC adapter: Chicony A17-230P1A
-            - AC power cord type: IEC C13
+    - 230W (19.5V, 11.8A) DC-in port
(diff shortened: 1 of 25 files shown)


XV-02:
Is this expected to behave broadly as the x11gestures extenions, that is:
- four-touch up/down - switch between workspaces
- four-touch left - open workspaces view
- four-touch right - open application view
or is there other functionality that it should introduce?

jacobgkau:
This isn't quite what we talked about. This still leaves the AC adapter lumped in as a subitem of the DC-in jack specifications. Do you mind if I do the rest, or do you want to discuss it some more?

For reference:

image

The first section is about the port on the laptop; the second section is about the AC adapter; the third section is about the battery. Three separate components that are all connected as parts of the power subsystem.

leviport:
Ah I see what you mean, I didn't realize the first part was describing the laptop's port. That makes a lot of sense though.

>Do you mind if I do the rest

Not at all, thanks for taking over!

jacobgkau:
> I didn't realize the first part was describing the laptop's port.

I've changed `DC-in power` to `DC-in port` on that line to address this confusion.

Samples of the new format:

darp7:
image

kudu6:
image

bonw14:
image

galp5:
image

jacobgkau (approved):

storyaddict (approved):
Changes look good. Clean and clear. Thanks @jacobgkau and @leviport

PR #136, opened 2022-06-28
Add darp8 & lemp11

jacobgkau:
This PR will add the Darter Pro 8 and the Lemur Pro 11 together, since they are very similar machines with similar updates and are being released very close together.

I'm opening this as a draft. I'm still working on the lemp11 section; darp8 is just about done (still needs proofreading.)


jacobgkau:
> * For replacing the cooling system on the darp8 the screws are removed in order but on the lemp11 they are removed in reverse order, is there a reason for this?

They have historically followed what the upstream Clevo service manual has recommended for each individual model. In truth, it probably doesn't matter much. This PR is currently just carrying forward the recommended orders from the darp7 and lemp10.

> The lemp11 has a section for removing the speakers can we have that in the darp8 as well or is it too difficult?

The Lemur Pro uses a non-standard mounting system with the rubber nubs. The Darter Pro uses regular Phillips screws and it should be fairly self-explanatory how to remove the speakers, plus this is not a very common part to replace as far as I'm aware. I'm not sure we need to set the precedent of including this section for every single model, although if the support team is seeing a large volume of speaker replacements happening, it could be done.

ahoneybun (dismissed):

ahoneybun:
In that case I've changed my review as I believe these are good to go and those were more questions then anything else.

src/models/lemp11/README.md
```diff
@@ -0,0 +1,57 @@
...
+    - GPU: Intel Iris Xe Graphics
+        - GOP driver is recommended, VBT is provided
+    - eDP display: 14.0" 1920x1080@60Hz LCD
+        - LCD panel: Innolux N140HCE-EN2 (or equivalent)
+    - External video outputs:
+        - 1x HDMI 2.1
+        - 1x DisplayPort 1.4 over USB-C
+- Memory
+    - Channel 0: 8-GB on-board DDR4
+        - 4x [Samsung K4AAG165WA-BCWE](https://www.samsung.com/semiconductor/dram/ddr4/K4AAG165WA-BCWE/) (2GB each)
```
crawfxrd, line 29: Schematics are wrong. This part is actually P4AAF165WA-BCWDE, per the SPD data provided.

jacobgkau: @crawfxrd Thanks for the correction! I can't seem to find the new part on [Samsung's website](https://semiconductor.samsung.com/dram/) (their part numbers seem to usually start with `K`), Google's returning zero results, and Bing's not returning any relevant results. Do you know if these (P4AAF165WA-BCWDE) are still Samsung brand and if there's a webpage for them anywhere?

crawfxrd: Proprietary firmware says it's still Samsung, and the part number is pretty close (K -> P, G -> F, +D).

Here is the SMBIOS data:

```
Handle 0x0021, DMI type 17, 92 bytes
Memory Device
	Array Handle: 0x0020
	Error Information Handle: Not Provided
	Total Width: 64 bits
	Data Width: 64 bits
	Size: 8 GB
	Form Factor: SODIMM
	Set: None
	Locator: Controller0-ChannelA
	Bank Locator: BANK 0
	Type: DDR4
	Type Detail: Synchronous
	Speed: 3200 MT/s
	Manufacturer: Samsung
	Serial Number: 00000000
	Asset Tag: 9876543210
	Part Number: P4AAF165WA-BCWDE    
	Rank: 1
	Configured Memory Speed: 3200 MT/s
	Minimum Voltage: Unknown
	Maximum Voltage: Unknown
	Configured Voltage: 1.2 V
	Memory Technology: DRAM
	Memory Operating Mode Capability: Volatile memory
	Firmware Version: Not Specified
	Module Manufacturer ID: Bank 1, Hex 0xCE
	Module Product ID: Unknown
	Memory Subsystem Controller Manufacturer ID: Unknown
	Memory Subsystem Controller Product ID: Unknown
	Non-Volatile Size: None
	Volatile Size: 8 GB
	Cache Size: None
	Logical Size: None
```

In fact, it looks like the lemp10 also is a different part according to the SPD data (M471A1G44AB0-CWE).

leviport:
I think something went wrong with combining the intro and model list pages. All of the links on the home page 404 because they don't point into the `models` directory (ex: `http://localhost:3000/addw2/README.html`).

jacobgkau:
> I think something went wrong with combining the intro and model list pages. All of the links on the home page 404 because they don't point into the `models` directory (ex: `http://localhost:3000/addw2/README.html`).

Interesting, the file was still in the `models` directory, but since mdbook treated that as the fallback homepage, it rendered the links as if it was in the document root.

I've moved the homepage from within the models directory back up to the actual document root, and updated the links to compensate for that. One remaining quirk is that "Current Models" and "Previous Models" in the sidebar aren't active when loading the homepage from the default `/` path (only when loading via `/README.md`, so when they're clicked), but since they both light up when either one is clicked anyway, that's not really important for navigation.

src/models/lemp11/README.md
jacobgkau: > In fact, it looks like the lemp10 also is a different part according to the SPD data (M471A1G44AB0-CWE).

I physically confirmed the lemp10's part number with a photo here: https://github.com/system76/tech-docs/pull/49#issuecomment-747610268 

The photo shows lemp10 had K4AAG165WA-BCWE. Are you sure this isn't a matter of a more internal part number vs. an overarching model number that Samsung would advertise?

That PR did say the printed part number (which is what we went with) was different from what was in the Clevo manual, so maybe getting a photo of the lemp11's RAM chips on the back of the motherboard would be a good idea.)

src/models/darp8/repairs.md
```diff
@@ -0,0 +1,211 @@
...
+**Difficulty:** Easy <span style="color:green;">●</span>  
+
+### Steps to remove the bottom cover:
+
+1. Place the machine lid-side down.
+    - Use a soft surface (such as a towel) to avoid scratches.
+2. Remove the 16 bottom panel screws:
+    - 12 perimeter screws (thin), highlighted green below.
+    - 2 keyboard screws (thicker), highlighted yellow.
+    - 2 inner hinge screws (thickest), highlighted red.
```
leviport, line 29: The hinge and keyboard screws are the same size on darp8

src/models/darp8/repairs.md
```diff
@@ -0,0 +1,211 @@
...
+2. Unscrew the retainer screw opposite the M.2 slot.
+    - SSD-1 is closest to the battery (pictured on the left.)
+    - SSD-2 is closest to the outer edge of the machine (on the right.)
+
+![M.2 slots](./img/m2-slots.webp)
+
+3. Remove the existing M.2 drive by pulling it out of the slot.
+4. Insert the new M.2 drive into the slot and hold it in place.
+5. Replace the retainer screw.
+6. If a square brace was present on the old drive, transfer it to the new drive.
```
leviport, line 73: Does "square brace" refer to the thermal pad? I guess it's a pretty minor point, but it sounds a little odd to call it a brace.

src/models/darp8/repairs.md
jacobgkau: I wasn't 100% sure looking at them in person, must have been my imagination making me think they were still different. I'll update.

src/models/darp8/repairs.md
jacobgkau: Yeah, I'll switch the wording to call it a thermal pad like we've started doing for other models. With some models it's been foam that seems more like a mechanical thing, other times it's the oily rubber more obviously for thermals.

src/models/darp8/repairs.md
jacobgkau: I didn't skimp out on updating the colors in the image, but upon further consideration, the keyboard screws are obviously good to be able to locate easily (even though it's called out again in the keyboard section), and it could be useful to know that the remaining two are for the hinges. With that in mind, I kept them colored differently, and just updated the wording to describe both as "thick" compared to the perimeter screws being "thin."

Let me know if you disagree, I can still update the colors if we want the four thicker ones to be colored the same despite their differing functions.

leviport: Yeah that makes sense to me

src/models/lemp11/README.md
crawfxrd: Can't see shit in the service manual, and I'm not taking it out, so I don't know what is actually attached to the board.

Considering I can make it say anything in the SPD data, it might be an internal/config thing.

jacobgkau: @leviport took apart our lemp11 and got a picture:

PXL_20220701_192934339

Looks like K4AAG165WA-BCWE is correct for lemp11, same as lemp10.

leviport (approved):
Looks great!

PR #137, opened 2022-07-05
Correct lemp11 CPU specs

leviport:


jacobgkau (approved):
Confirmed these are correct for lemp11/lemp11 doesn't use the same CPU models as darp8.

ahoneybun (approved):
Links are good and it works correctly live.

PR #138, opened 2022-07-12
Add Launch Lite docs

leviport:


leviport:
I still need to update most of the pictures once I'm back in the office and have access to a Lite, but this should be most of what is needed for the text.

leviport:
Thanks @jacobgkau, your improvements look great!

src/models/launch_1/README.md
```diff
@@ -16,7 +16,11 @@ The System76 Launch is a configurable keyboard with the following specifications
   - N-key rollover
 - Sockets and Switches
   - Kailh MX hot swap sockets
-  - Kailh Box Jade or Box Royal switches
+  - Default switch options:
+    - NovelKeys x Kailh Box Jade
```
thomas-zimmerman, line 20: Why the callout that 2 of the options are "NovelKeys x Kailh Box XXX" keys?  I may make more sense to follow the Configure and Buy description and have the sound profile:
 * Kailh Box Jade - Clicky
 * Kailh Box royal - Tactile
 * Kailh Box Silent Pink - Silent Linear
 * Kailh Box Silent Brown - Silent

src/models/launch_lite_1/README.md
```diff
@@ -0,0 +1,44 @@
...
+  - [Open-source milled chassis design](https://github.com/system76/launch/tree/master/chassis/launch-lite)
+  - Magnetically-attachable 15-degree angle lift bar
+- Electronics
+  - [Open-source PCB design](https://github.com/system76/launch/tree/master/pcb-lite)
+  - Individually addressable RGB LED backlighting
+  - N-key rollover
+- Sockets and Switches
+  - Kailh MX hot swap sockets
+  - Default switch options:
+    - NovelKeys x Kailh Box Jade
```
thomas-zimmerman, line 20: As on the Launch, I think we should follow the sales description:
 * Kailh Box Jade - Clicky
 * Kailh Box royal - Tactile
 * Kailh Box Silent Pink - Silent Linear
 * Kailh Box Silent Brown - Silent

src/models/launch_lite_1/repairs.md
```diff
@@ -0,0 +1,158 @@
...
+![Launch Lite with all switches removed](./img/all-switches-removed.webp)
+
+2. Turn the keyboard over and remove the four bottom panel screws.
+    - The screws are set into adapters that may or may not come out with the screws.
+
+![Bottom panel screws](./img/bottom-panel-screws.webp)
+
+3. Remove the bottom panel.
+    - If the magnetic lift bar is installed, it can be used as a handle.
+    - A small plastic spacer (highlighted cyan below) may come loose; if it does, place it back in the middle PCB hole.
```
thomas-zimmerman, line 123: The center plastic spacer (highlighted in cyan below) is not attached to the PCB. Remove and save to replace in center hole on installation of the PCB (new or old).

thomas-zimmerman (dismissed):
I've left some comments that I think are improvements but this is great!

src/models/launch_1/README.md
leviport: It's explained in the commit message

src/models/launch_lite_1/repairs.md
leviport: It's supposed to snap into the hole, it's just somewhat easy to dislodge.

jacobgkau: @leviport For what it's worth, it did not "snap" at all on the unit I had for photos. I could tell from its shape that it was supposed to snap (after you explained it to me), but either the hole in the PCB was too large or the plastic piece was too small. It sounds like this may be a wider issue than just a lab unit quirk if Thomas also knew about it?

I do think the current description covers the case of the spacer not being attached and is simpler than the suggested update. The current description doesn't say anything about snapping. (The second sentence in the suggested update is missing a subject, too.)

src/models/launch_1/README.md
jacobgkau: A web search for "Box Jade" or "Box Royal" show some websites simply referring to them as "NovelKeys Box Jade" / "NovelKeys Box Royal," so including NovelKeys in the specs here does make sense. At least a couple of other websites (KBDfans and Divinikey, from the first page of results) use the "NovelKeys x Kailh" format.

If we want to add the sound profile, I would recommend placing them in parenthesis instead of separating them with a hyphen, or having sub-bullets like `Sound Profile: Clicky` under each switch option.

src/models/launch_lite_1/repairs.md
leviport: I'll consult Mechanical Engineering about it. There were some part substitutions being made for that spacer, and the keyboard I handed you was a pre-production unit, so the diameter of the spacer probably increased a bit to stay put a little better.

src/models/launch_1/README.md
leviport: What I _really_ wanted to do was to link to a spec sheet for each switch, but Kailh's datasheet is pretty heavy, and Novelkeys's page on the switches are also product sales pages, not purely documentation pages. Regardless, it's pretty common for keyboards to simply list switch types without giving sound/tactility details, so I think it's fine the way it is.

src/models/launch_lite_1/external-overview.md
```diff
@@ -0,0 +1,12 @@
...
+## Top overview:
+
+![Top Overview](./img/overhead.webp)
+
+## Back overview:
+
+![Back Ports](./img/ports-back.png)
+
+## Box contents overview:
+![Launch Box Contents](./img/launch-lite-box-contents.png)
```
Ampersandstorm, line 12: Can these accessories be replaced or purchased from System76?

Ampersandstorm (approved):
Looks good. I just had one question about replacing accessories.

src/models/launch_lite_1/external-overview.md
jacobgkau: I'm pretty sure the tool and cables can be purchased/replaced at the discretion of support, probably the lifter bar too. The extra keys may depend on what we currently have on hand since we purchase the keys as complete sets from our supplier.

thomas-zimmerman: It's usually be at discretion of production an availability of parts. Individual key caps are not stocked so purchase would be for full set, for example. The lead time on some of the parts for the launch is limiting.

Ampersandstorm: Do you think it's worth mentioning that many of these accessories can be sourced by the user if lost or broken (I assume the USB cables are not a unique specification, and that none of these tools are proprietary). I know key pullers are easy to come by, but I'm not sure about the switch puller. Maybe there's a more appropriate place to mention it, but I couldn't find anything on the product page or in the support docs.

src/models/launch_lite_1/repairs.md
leviport: Alright, I talked it over with John. I guess the plastic piece is a _spacer_ that doesn't really snap into the PCB very well. I think the wording here is probably good though.

src/models/launch_lite_1/external-overview.md
leviport: How about:

> - Contact [System76 Support](https://support.system76.com/) to purchase replacement accessories. Availability may vary.

src/models/launch_1/README.md
leviport: I'm going to resolve these switch comments if we're happy with it the way it is. Feel free to unresolve if I'm jumping the gun.

Ampersandstorm (approved):
Looks good

PR #140, opened 2022-08-10
Add oryp9

jacobgkau:
This will add the Oryx Pro 9 to tech docs.


src/models/oryp9/repairs.md
```diff
@@ -0,0 +1,252 @@
...
+![Thermal paste removal](./img/thermal-paste-removal.webp)
+
+8. If you are replacing the heat-conductive metal, remove the plastic covers from the GPU heatsink and the CPU chip.
+    - A fingernail or a thin plastic tool can be used to pick a side of the plastic cover up from the CPU chip/GPU heatsink.
+    - After removing the plastic covers, wipe away any remaining heat-conductive metal residue from the heatsink, dedicated GPU die, and main CPU die.
+    - Some adhesive residue may remain on the GPU heatsink and CPU chip.
+
+![Heat-conductive metal removal](./img/heat-conductive-metal-removal.webp)
+
+9. Apply the new heat-conductive metal to the CPU chip and the GPU heatsink, matching the orientation of the plastic coverings with the diagonal corners of their respective surfaces.
```
leviport, line 182: >Apply the new heat-conductive metal to the CPU chip and the GPU heatsink

Shouldn't we advise that they put the metal on the GPU itself rather than the heatsink? Otherwise it'd be difficult to hold onto the heatsink while putting it back together, and it could be misaligned so that the GPU die doesn't sit in the cutout on the plastic part.

src/models/oryp9/repairs.md
```diff
@@ -0,0 +1,252 @@
...
+6. Remove the heatsink/fans from the case, being careful not to bend the heatsink pipes. It may take some pressure to break the seal of the thermal paste.
+7. Using a paper towel, remove the existing thermal paste from the integrated CPU die, the border of the CPU chip, the CPU heatsink, the border of the GPU die, and the GPU heatsink.
+    - A small amount of rubbing alcohol can be used if the old paste is dried or difficult to remove.
+    - The heat-conductive metal may have some holes where pieces stuck to the die or heatsink.
+    - If you are not replacing the heat-conductive metal, do not pick or wipe the heat-conductive metal off of the main CPU die or the GPU heatsink.
+
+![Thermal paste removal](./img/thermal-paste-removal.webp)
+
+8. If you are replacing the heat-conductive metal, remove the plastic covers from the GPU heatsink and the CPU chip.
+    - A fingernail or a thin plastic tool can be used to pick a side of the plastic cover up from the CPU chip/GPU heatsink.
```
leviport, line 176: There should be a warning about not getting the metal stuff anywhere on the motherboard, but I'm not sure whether it works better here or closer to the top, near the room temperature warning.

src/models/oryp9/repairs.md
jacobgkau: No, the metal comes off with the heatsink when removing the heatsink and the guides to place the metal/plastic cover are located on the heatsink for the GPU.

jacobgkau: To clarify, there is only adhesive on one side of the plastic cover. That side needs to be facing the heatsink in order to be in the correct orientation. So it needs to be stuck onto the heatsink.

src/models/oryp9/repairs.md
jacobgkau: I'll copy some of the warning under Step 1 down here as well.

src/models/oryp9/repairs.md
leviport: Ah I gotcha, that makes sense. I misunderstood how that worked.

src/models/oryp9/repairs.md
jacobgkau: Let me know how it looks with [a7ca2dd](https://github.com/system76/tech-docs/pull/140/commits/a7ca2dd58fb26ddd7e6b11427ea0429ccc6fb814). I almost put "if powered on at room temperature," but I've personally killed a board without powering it on (and the thermal instructions don't include unplugging the battery), so this seems safer.

leviport: Looks good to me!

jacobgkau:
This PR also adds the BIOS chip form factor (SOIC-8 or WSON-8) to all models.

leviport (approved):

jacobgkau:
If support finds any problems, they can be addressed with a follow-up PR or within the upcoming galp6 branch.

PR #141, opened 2022-08-17
Adds lemp9 Wi-Fi card

ahoneybun:
This PR does the following:

- Adds the model information for the Wi-Fi card that shipped in the lemp9


ahoneybun:
It looks like it can also use the Intel Wi-Fi 6 AX200/AX201 which is in the lemp10/11 do we also want to add that?

jacobgkau:
> It looks like it can also use the Intel Wi-Fi 6 AX200/AX201 which is in the lemp10/11 do we also want to add that?

Yes, we sold the lemp9 with the option of standard WiFi (9560) or WiFi 6 (AX20x).

jacobgkau (approved):

PR #145, opened 2022-09-01
Correct darp8 charger

leviport:


jacobgkau (approved):
Confirmed the wattage was wrong comparing with the upstream Clevo service manual, the charger model would need to be updated to one with the appropriate wattage (and it is).

PR #146, opened 2022-09-07
gaze17: Updates for 3050x GPUs, 15" display models

jacobgkau:
This updates the specs and adds an additional right port diagram to specify that the 3050/3050 Ti models do not have Thunderbolt.

It also adds 15" dimensions that were missing, adds the keyboard/touchpad overview image for 15" and 17", and changes the external overview headings to match newer models' sections.


Ampersandstorm (approved):
Looks good

PR #151, opened 2022-09-30
Correct battery size

ahoneybun:


n3m0-22 (approved):
Assuming that is the correct picture for this model this should be good.

ahoneybun:
@n3m0-22 it looks like the galp5 had a 49Wh battery so it was copied over from there based on this photo:

https://tech-docs.system76.com/models/galp5/img/components-highlighted.jpg

PR #152, opened 2022-10-05
Add missing GB in tech specs

ahoneybun:


n3m0-22 (approved):
Looks like you found them all.

PR #153, opened 2022-10-07
Thelio b4.0

jacobgkau:
This will add the Thelio b4.0 to tech docs.


jacobgkau:
This is starting to feel like I could keep making revisions forever, but it's probably ready to review at this point.

I had to digitally alter the back overview to correct the inaccurate etched port labels (HDMI & DP version numbers), which I don't like doing. I don't know if we have an ETA on a corrected chassis being produced and available for a photo, though (and I believe the normal lights are gone from the photo room this week.) We can potentially swap that photo out later if what's there now doesn't look too bad.

One other item of note, I went with just `thelio-b4` for the model of this section as opposed to `thelio-b4.0` since the `.0`/`-0` formatting is a little inconsistent and we wouldn't typically do a new Tech Docs section for a minor revision anyway. If this leads to confusion, we can add `.0` or `-0` back in the future to clarify. I did add a note to the top of the repairs page about the revision similar to the ones we use for laptop pages regarding their screen size variations.

src/models/thelio-b4/external-overview.md
```diff
@@ -0,0 +1,18 @@
+# Thelio (External Overview)
+
+## Back overview:
+
+![Back Ports](./img/ports-back.webp)
```
leviport, line 5: I believe the ethernet port should be labeled `2.5Gb/s` or `2.5Gb` rather than `2.5GiB`. GiB is gibibyte, and gibibyte != gigabyte != gigabit.

src/models/thelio-b4/repairs.md
```diff
@@ -0,0 +1,455 @@
...
+1. Follow the steps above to [remove the top case](#removing-the-top-case).
+2. Unscrew the two screws securing the drive bay's cover, highlighted green below.
+
+![2.5" drive cover and 2.5" drive screws](./img/25-drive-screws.webp)
+
+3. If you are adding a new drive, pop out a black plastic ring on the top crossbar and slide out four screws (per drive).
+4. Insert four screws into each 2.5" storage drive you wish to install.
+5. Slide each 2.5" drive into one of the slots leading to the Thelio-IO board.
+
+![2.5" drive installation](./img/25-drive-installation.webp)
```
leviport, line 75: It looks like this image is missing from the img directory. I'm only seeing the description text.

src/models/thelio-b4/repairs.md
```diff
@@ -0,0 +1,455 @@
...
+
+Thelio B4 supports up to 64GB (2x32GB) of RAM. The RAM sticks are DDR5 DIMMs (non-ECC) running at a speed of up to 5600MHz. If you've purchased new RAM, need to replace your RAM, or are reseating your RAM, follow these steps.
+
+**Tools required:** Cross-head (Phillips) screwdriver  
+**Time estimate:** 15 minutes  
+**Difficulty:** Medium <span style="color:orange;">●</span>
+
+### Steps to replace the RAM:
+
+1. Follow the steps above to [remove the top case](#removing-the-top-case) and [remove the CPU shroud]().
```
leviport, line 182: CPU shroud link is empty

leviport (changes requested):
I found a couple small bugs, but this is dynamite overall. I'll approve once those are taken care of

src/models/thelio-b4/external-overview.md
jacobgkau: You're right, I used the wrong abbreviation. [Microsoft's styling guide](https://learn.microsoft.com/en-us/style-guide/a-z-word-list-term-collections/term-collections/bits-bytes-terms) and the motherboard manual use `Gbps`, so I'll go with that. (Intel ARK uses `GbE`.)

jacobgkau: Fixed with https://github.com/system76/tech-docs/pull/153/commits/21c55ff87f2b786a5bf7bbd8cdda2f6d69b43dd0.

src/models/thelio-b4/repairs.md
jacobgkau: Fixed with https://github.com/system76/tech-docs/pull/153/commits/6603dd6300d60022bf55253eb703098b484d37a4.

src/models/thelio-b4/repairs.md
jacobgkau: I forgot that photo had a focus issue. I retook it today and will drop it in once I can get it off of my camera.

leviport (approved):
Looks great, nice work! I'll let you merge this once you're ready.

src/models/thelio-b4/repairs.md
ahoneybun, line 75: I believe this should be: 25-drive-installed.webp

ahoneybun (changes requested):
Some of the image filenames seem to be incorrect causing them to not load.

ahoneybun (approved):
This is just fantastic! I love the section for the Thelio Io board wiring!

PR #155, opened 2022-11-15
Fix NVMe Gen number

ahoneybun:


n3m0-22 (approved):

jacobgkau (changes requested):
The readme (spec list) indicates that Gen 2 is the correct number. If this is actually wrong, then the readme would also need to be updated.

ahoneybun:
> The readme (spec list) indicates that Gen 2 is the correct number. If this is actually wrong, then the readme would also need to be updated.

I put a Gen 3 drive in the lab unit the other day for testing the backlight issue on this model and it worked without issue. I'll update the README as well.

jacobgkau (changes requested):
We already tested speeds to confirm that Gen 2 is correct: https://github.com/system76/tech-docs/pull/34 

This update is inaccurate.

PR #164, opened 2023-02-23
Pangolin 12

jacobgkau:
This adds the Pangolin 12 (pang12) to tech docs.

Information we typically include that is currently missing -- please let me know if we can get these:
- The BIOS flash chip model number & location (specs, internal overview).
- The LED meanings (external overview).

Also, the external overview keyboard shortcut icons are slightly inaccurate since they're still using the Clevo graphics. Let me know if we have graphics that more closely resemble the Emdoor keyboard.

I went ahead and added a lid photo to the External Overview, which we should have for all laptops moving forward. This is something that's been requested since the lid/logo is often not shown on our sales pages, but we haven't had them because the marketing team doesn't always take photos of the lid (since they're not needed for the sales pages.) However, since QA is doing photos for the Thelio external overviews now, I don't see why we can't also do extra photos for laptop externals.


ahoneybun:
@jacobgkau thanks for adding this:

"During normal operation, only the left fan will spin. The right fan is reserved in the embedded controller firmware for cooling a dedicated GPU, which the Pangolin 12 does not have."

The fan behavior seems to be confusing for customers.

ahoneybun (approved):
This looks great as usual! Those updated key combo logos would be good but this looks good to go to me.

jacobgkau:
Added BIOS chip, RAM chip, and LED info; updated the keyboard shortcuts to use the new symbols.

(The screen-printed LED label color happens to be visible against all four of the website background colors, so I set all of the keyboard and LED symbols to that color.)

src/models/pang12/repairs.md
```diff
@@ -0,0 +1,174 @@
...
+2. Pull the M.2 access door off, starting from the top (larger side).
+    - A flat, plastic tool can be used to pry the edge of the door up.
+    - If the door is difficult to remove, the M.2 drives can alternatively be accessed by [removing the bottom cover](#removing-the-bottom-cover).
+3. Unscrew the retainer screw opposite the M.2 slot.
+
+![M.2 slots](./img/m2-slots.webp)
+
+4. Remove the existing M.2 drive by pulling it out of the slot.
+5. Insert the new M.2 drive into the slot and hold it in place.
+    - If a slot is being used for the first time, the spare thermal pad (which is included in the Pangolin's box) can be applied to the motherboard before placing the drive down on top of it.
```
leviport, line 36: The thermal pad doesn't really go on the motherboard here, since the M.2 connectors are at the very edge of the board. It goes on the little silver heatsink thing, however you want to describe that.

jacobgkau: [8fa492a](https://github.com/system76/tech-docs/pull/164/commits/8fa492a9a24f64c5bc89f2bb00bd9f4aaa6a0c13) - `silver backplate`, how's that?

leviport: Sounds great!

leviport (approved):

ahoneybun (approved):

PR #166, opened 2023-02-27
Fixes for legacy Thelio Major section

jacobgkau:
The primary purpose of this PR is to eliminate the top-level `pictures` directory in the mdbook source. This directory was not used as it was originally intended, and is not necessary.

(We've established by now that each section/model has its own image directory containing all of the images for that specific model. Reusing photos between sections is highly discouraged, and when we do, the image is simply copied between sections so each section can be independently edited, archived, etc. without worrying about the others.)

In order to do this, the images from this directory were moved into the legacy Thelio Major b1/b2/r1/r2 section (which is the only section that used any images from this directory.) I moved these one at a time, renaming them with more concise and descriptive filenames where appropriate. As a side effect, this PR also removes various images that were not actually used anywhere in the mdbook or on the website (this saves bandwidth and time when cloning the repo.)

I also made a few miscellaneous fixes listed in the second commit. This was not a full edit/revision of the section, just some obvious items that I noticed while updating the image links.


jacobgkau:
Yeah, I didn't look as closely at the ones already in the correct folder. Those two were redundant.

Renamed the internal & external overview photos to be clearer and more in line with the current naming scheme, removed that redundant photo along with another one that I found (the M.2 slot diagram was duplicated as `thelio-major_diagram_config.png`).

leviport (approved):

PR #168, opened 2023-03-15
Add part numbers for darp8

ahoneybun:
This PR adds the top case and bottom panel part numbers for the darp8.


jacobgkau (approved):
I've added part numbers for the LCD back cover (lid) and front cover (bezel) based on the Clevo service manual.

I see no reason not to list these if we have them. We might need to revisit the format in the future if we find some models have a lot of alternative compatible part numbers for the same part.

PR #169, opened 2023-03-27
Numlock + Fan

ahoneybun:
This adds the Fan control like other models and about NumLock feature of this laptop. This enables a NumPad like section of the keyboard.

### DIFF
--- a/src/models/galp4/external-overview.md
+++ b/src/models/galp4/external-overview.md
@@ -35,7 +35,7 @@ The Galago Pro has the following actions available using the Fn and Function key
 |![Fn-F11](./img/fn-f11.png)|Fn+F11  |Toggle airplane mode               |
 |![Fn-F12](./img/fn-f12.png)|Fn+F12  |Suspend                            |
 |![Fn-F12](./img/fn-dia.jpg)|Fn+`    |Play/Pause                         |
-|Ins                        |Fn+Ins  |Enable Numpad                      |  
+|Ins                        |Fn+Ins  |Toggle numpad                      |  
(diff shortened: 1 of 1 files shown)


src/models/galp4/external-overview.md
```diff
@@ -35,6 +35,8 @@ The Galago Pro has the following actions available using the Fn and Function key
 |![Fn-F11](./img/fn-f11.png)|Fn+F11  |Toggle airplane mode               |
 |![Fn-F12](./img/fn-f12.png)|Fn+F12  |Suspend                            |
 |![Fn-F12](./img/fn-dia.jpg)|Fn+`    |Play/Pause                         |
+|Ins                        |Fn+Ins  |Enable Numpad                      |
```
jacobgkau, line 38: Is the numpad a toggle (press/release the shortcut to turn it on, press/release again to turn it off), or do you have to hold down Fn+Ins while using the numpad?

ahoneybun: It is a toggle similar to the other functions.

jacobgkau: Cool, I'd like to call it a toggle to match the other lines and clarify that it also disables and doesn't just enabled, then, if that's okay. Made the change with 4e8a06b.

jacobgkau (approved):

src/models/galp4/external-overview.md
ahoneybun: Do we want to change the verbiage to Toggle Numpad thinking about it now.

jacobgkau: Yep, that's exactly what I did with that commit.

PR #170, opened 2023-03-28
Customer feedback on battery replacement

thomas-zimmerman:
Add some additional advise on removing the battery and how to connect the battery again.


src/models/lemp11/repairs.md
```diff
@@ -188,9 +188,12 @@ The battery provides primary power whenever the system is unplugged.
...
+    - The battery is secured to foam standoffs with double-sided tape, with tabs of the adhesive that extend from under the battery. Pulling the tabs will remove most of the adhesive. 
+    - For the tabs sticking out on the left side, gently removing the speaker will give you more room to access the adhesive tabs. 
     - A flat plastic object (such as a credit card) can be used to separate the tape from the battery.
     - If any tabs from the tape are visible around the perimeter of the battery, do not push them back under the battery.
         - Hold the tabs down while inserting the plastic tool to separate the tape and foam from the battery.
         - Alternatively, pull the tabs up with the battery to separate the tape and foam from the chassis.
+    - Leave the foam backing in place for the replacement battery to rest on.
 4. Put the battery back (or put your new battery in its place) and plug it back into the motherboard.
+    - You may need to peal back the tape that holds the battery connection cable to tuck the cable under the battery when attaching back to the white connection on the main logic board.
```
leviport, line 199: s/peal/peel

leviport: >the tape that holds the battery connection cable to tuck the cable under the battery when attaching back to the white connection on the main logic board

This is hard to parse, could we reword it a bit to make it clearer?

leviport:
I believe these sections are pretty much identical between lemp9, lemp10, and lemp11. Should we make similar changes in lemp11 and lemp9 as well?

src/models/lemp11/repairs.md
jacobgkau: It also shouldn't be called `logic board` when the previous sentence used `motherboard`.

jacobgkau:
@thomas-zimmerman What is the actual recommended method you want the instructions to say? We have:

- (existing) Wedge a plastic tool between the battery/tape to separate them.
- (existing) Pull the tabs up to separate the tape and battery from the chassis.
- (new) "Pull the tabs to remove most of the adhesive" is what you originally added here-- I assume you mean pulling it outward, rather than up away from the chassis as was already suggested?

I think this is too many alternatives-- we're basically not giving people instructions if we just list every possible way to go about it.

thomas-zimmerman:
> @thomas-zimmerman What is the actual recommended method you want the instructions to say? We have:
> 
> * (existing) Wedge a plastic tool between the battery/tape to separate them.
> * (existing) Pull the tabs up to separate the tape and battery from the chassis.
> * (new) "Pull the tabs to remove most of the adhesive" is what you originally added here-- I assume you mean pulling it outward, rather than up away from the chassis as was already suggested?
> 
> I think this is too many alternatives-- we're basically not giving people instructions if we just list every possible way to go about it.

They may need to use all of these alternate ways to remove the adhesive. The pull tabs can break and leave a decent amount of adhesive on the bottom of the battery, and you will be using a plastic tool to try and remove the excess. 

Good catch, I do mean that they will need to pull outward from the battery, not away from the chassis.

jacobgkau:
I'm good with leaving a form of all three in, but two of them (as currently worded) are mutually exclusive. Before the tabs break, you have to pull them out or up, you can't really do both. Unless we just want to say "upwards and outwards" for both at once.

thomas-zimmerman (commented):
Yes, I do think that is an improvement. No odd formatting when looking at the live view in mdbook. I'd approve the PR but not able as the author.

src/models/lemp11/repairs.md
```diff
@@ -187,14 +187,15 @@ The model number for the Lemur Pro 11's battery is `L140BAT-4`, and the original
...
 
 ![Main battery](./img/battery.webp)
 
 3. Remove the battery.
     - The battery is secured to foam standoffs with double-sided tape.
-    - A flat plastic object (such as a credit card) can be used to separate the tape from the battery.
-    - If any tabs from the tape are visible around the perimeter of the battery, do not push them back under the battery.
-        - Hold the tabs down while inserting the plastic tool to separate the tape and foam from the battery.
-        - Alternatively, pull the tabs up with the battery to separate the tape and foam from the chassis.
+    - If any tabs from the tape are visible around the perimeter of the battery, pull them outwards and upwards to separate the battery from the chassis. Remove as much of the tape as possible.
```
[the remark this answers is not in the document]
jacobgkau: That sounds essentially like the keyboard adhesive. I can change the advice to that.

leviport (dismissed):

jacobgkau:
@thomas-zimmerman FYI, this has changed somewhat significantly again since your previous acknowledgement. I'm assuming this is still fine with you as well?

jacobgkau:
Just had to correct the number of the final step since I'd added one.

thomas-zimmerman:
I'm good with the additional changes!

PR #173, opened 2023-04-21
Copy image from mira-r3 for the Thelio Io button

ahoneybun:
This image is missing from major-r3 tech docs so copied it from mira-r3.


jacobgkau (changes requested):
We shouldn't use a mira-r3 photo in the major-r3 section. I'll check if I have a photo for this, or try to get one if not.

leviport:
Here's a quick one, just for comparison, of major-r3. Since the thelio-IO and drive cage mounting points are pretty much the same between mira-r3 and major-r3, I don't think the difference is enough for it to matter.

IMG_20230424_174624_835

jacobgkau:
I found the photos I took for this purpose while shooting.

jacobgkau (approved):

PR #177, opened 2023-05-24
Spec update for serw13

thomas-zimmerman:
Don't advertise DP support on non-TB port if it's not reliable

### DIFF
--- a/src/models/serw13/README.md
+++ b/src/models/serw13/README.md
@@ -58,7 +58,7 @@ The System76 Serval WS is a laptop with the following specifications:
         - Supports USB-PD (charging) when the system is powered off or suspended.
         - Supports DisplayPort over USB-C.
     - 1x USB 3.2 Gen 2 Type-C
-        - Supports DisplayPort over USB-C.
+<!---        - Supports DisplayPort over USB-C.--->
     - 2x USB 3.2 Gen 1 Type-A
 - Dimensions
     - 15": 2.49cm x 35.8cm x 24.0cm, 2.4kg


src/models/serw13/README.md
```diff
@@ -58,8 +58,8 @@ The System76 Serval WS is a laptop with the following specifications:
         - Supports USB-PD (charging) when the system is powered off or suspended.
         - Supports DisplayPort over USB-C.
     - 1x USB 3.2 Gen 2 Type-C
-        - Supports DisplayPort over USB-C.
+<!---        - Supports DisplayPort over USB-C.--->
```
leviport, line 61: I don't think this is the desired effect:


Probably better to just delete the text, rather than trying to comment it out. Commenting things out of markdown isn't usually a recipe for a good time.

src/models/serw13/README.md
```diff
@@ -58,8 +58,8 @@ The System76 Serval WS is a laptop with the following specifications:
...
     - 2x USB 3.2 Gen 1 Type-A
 - Dimensions
     - 15": 2.49cm x 35.8cm x 24.0cm, 2.4kg
-    - 17": 2.49cm x 39.6cm x 26.2cm, 2.8kg
\ No newline at end of file
+    - 17": 2.49cm x 39.6cm x 26.2cm, 2.8kg
```
leviport, line 65: Why is this in the diff? Can it not be?

leviport (changes requested):

src/models/serw13/README.md
thomas-zimmerman: Yes, that is not the desired change. Does it make sense to reject and close this PR and just start fresh? I don't think the change history *needs* a dead end change in there.

leviport: I probably would have just fixed this one, amended the commit, and force pushed, but a new PR is okay if that's what you prefer.

src/models/serw13/README.md
leviport: Web editor can be unpredictable in my experience

PR #178, opened 2023-05-24
Pull back advertised support for video over USB-C to only document this as working on the USB-C TB port.

thomas-zimmerman:
…is as working on the USB-C TB port.

### DIFF
--- a/src/models/serw13/README.md
+++ b/src/models/serw13/README.md
@@ -32,7 +32,7 @@ The System76 Serval WS is a laptop with the following specifications:
     - External video outputs:
         - 1x HDMI
         - 1x Mini DisplayPort 1.4
-        - 2x DisplayPort 1.4 over USB-C
+        - 1x DisplayPort 1.4 over USB-C
 - Memory
     - Up to 64GB (2x32GB) dual-channel DDR5 SO-DIMMs @ 5600 MHz
 - Networking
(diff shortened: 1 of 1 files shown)


leviport (dismissed):

jacobgkau:
@leviport Regarding the comments over at https://github.com/system76/tech-docs/pull/177#issuecomment-1561367114, if the port doesn't support DisplayPort, then it should never work. Just to be clear, the port physically supports it, but it's unreliable, so rather than worrying about fixing that in firmware/software, we're just saying it doesn't support it since the Thunderbolt port still does anyway?

jacobgkau:
Ref: https://github.com/pop-os/gnome-shell-extension-system76-power/pull/89.

I don't really like this documentation-wise, but agree we shouldn't list both ports as fully functional, and explaining a partially functional port could be tricky.

leviport:
We removed it from the product specs on the sales page, so this is presumably to bring parity between the two. You're right that it does support it, only it's pretty unreliable. I'll defer to your judgment on this though, so do whatever you think works best in this situation.

jacobgkau:
Do you think this draws too much attention?

image

I'm thinking of scenarios where a customer could be on a different distro without the dialogue prompt, and support or the customer is looking for information about the symptoms. Support or the customer may not even think about which port it's plugged into if we only have one listed as being compatible.

leviport:
I think that's good

jacobgkau (approved):

leviport (approved):

PR #181, opened 2023-06-01
Update GPU section for Oryx Pro models

ahoneybun:
This matches with the gaze18 tech-docs.


jacobgkau (changes requested):
The 11th gen CPU in the oryp8 has UHD Graphics, not Iris Xe Graphics.

Also, each of your lines listing the Intel graphics has an extraneous trailing space.

jacobgkau (approved):
Looks good after correcting the oryp8 iGPU from Iris Xe to UHD and removing the trailing spaces.

PR #186, opened 2023-06-14
serw13 (specs): Add max USB-C charge rate

jacobgkau:
This whole section is getting a little wordy, but I guess we have a lot of nuances to explain for this product.

image

### DIFF
--- a/src/models/serw13/README.md
+++ b/src/models/serw13/README.md
@@ -55,7 +55,7 @@ The System76 Serval WS is a laptop with the following specifications:
     - MicroSD card reader
 - USB
     - 1x USB Type-C with Thunderbolt 4
-        - Supports USB-PD (charging) when the system is powered off or suspended.
+        - Supports USB-PD (charging) up to 100W when the system is powered off or suspended.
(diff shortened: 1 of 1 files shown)


ahoneybun:
Do we want to move it like the galp7? 
Screenshot from 2023-06-14 10-45-32

leviport (approved):

jacobgkau:
> Do we want to move it like the galp7?

Fine, although the formatting's already different because we have periods here (which I think we should keep since the lines are so long.) Also, Thunderbolt implies DisplayPort support (we just have that line there because people who don't understand that still asked), so USB-PD is actually the more informative line, especially now that there's a further note. For the record, I think the reason it's in this order on other models is because the Graphics section comes before the Power section earlier in the page (on the serw13 page, we don't list USB-C charging under the Power section since it's not compatible when the system's powered on and is more of a bonus feature-- another reason why the USB-PD line was more important and thus came first.)

leviport (approved):

ahoneybun (approved):

PR #192, opened 2023-06-29
nebula49

jacobgkau:
This adds the external overview and assembly manual for the nebula49.

### DIFF
--- a/src/models/nebula49-1/assembly.md
+++ b/src/models/nebula49-1/assembly.md
@@ -1,6 +1,6 @@
 # nebula49 (Parts & Assembly)
 
-The nebula36 chassis is ready to be outfitted with standard personal computer components. If the system has already been built, ensure the system is powered off and all cables are unplugged from the motherboard, PCIe cards, power supply, and front I/O before opening the chassis.
+The nebula49 chassis is ready to be outfitted with standard personal computer components. If the system has already been built, ensure the system is powered off and all cables are unplugged from the motherboard, PCIe cards, power supply, and front I/O before opening the chassis.
 
 The preinstalled velcro strips are left partially unwrapped to aid in removal. When building the system, you can optionally wrap the velcro up the rest of the way.
 
@@ -21,14 +21,14 @@ The preinstalled velcro strips are left partially unwrapped to aid in removal. W
 
 ## Replacing the front accent strip:
 
-nebula36 includes a customizable accent panel on the front of the case, which can be swapped to change the case's look and feel. The instructions for swapping the accent are also available in [video form](https://system76.com/r/diyaccentpanel).
+nebula49 includes a customizable accent panel on the front of the case, which can be swapped to change the case's look and feel. The instructions for swapping the accent are also available in [video form](https://system76.com/r/diyaccentpanel).
 
 - **Tools required:** None  
 
 ### Steps to replace the front accent strip:
 
-1. Place the nebula36 on the edge of the desk so the front side is hanging off of the desk.
-    - The nebula36 can alternatively be lifted or tilted so the front of the computer is hovering above the desk.
+1. Place the nebula49 on the edge of the desk so the front side is hanging off of the desk.
+    - The nebula49 can alternatively be lifted or tilted so the front of the computer is hovering above the desk.
 2. Slide the accent strip down to unlock it.
     - The accent can be gripped at the bottom edge.
 3. Pull the accent strip off of the case, starting with the bottom edge.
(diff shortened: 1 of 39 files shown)


jacobgkau:
This will now also include a couple of minor fixes for the nebula19 and nebula36 sections.

When merging, please do not squash (use `Rebase and merge`, not `Squash and merge`).

leviport:
This review is kinda tricky since Github prevents me from commenting on lines that weren't changed in the PR. I guess I'll just do individual comments instead.

At the top of `src/models/nebula49-1/README.md` it looks like the External Overview and Parts & Assembly links are missing. Also, the name should be changed to `nebula49` to match 36 and 19



(still reviewing)

leviport:
Not sure where we landed with the optional second side intake fan, but if we're making that an option, then this note will need to be updated in `src/models/nebula49-1/assembly.md` on L55


(still reviewing)

leviport:
On L126 of `src/models/nebula49-1/assembly.md`, the link `[remove the CPU shroud](removing-the-cpu-shroud).` is missing a `#` so it 404's

(still reviewing)

src/models/nebula49-1/assembly.md
```diff
@@ -220,7 +216,7 @@ These instructions use the Noctua `NH-U12S` as an example.
 
 ## Installing the GPU:
 
-nebula36 supports GPUs up to 318.80mm long. For optimal cooling and ease of installation, GPUs longer than 308.80mm are not recommended.
+nebula49 supports GPUs up to 318.80mm long. For optimal cooling and ease of installation, GPUs longer than 308.80mm are not recommended.
```
leviport, line 219: GPU lengths need to be updated

src/models/nebula49-1/assembly.md
```diff
@@ -250,9 +246,9 @@ nebula36 is designed to work with an optional SATA backplane to allow for easy h
 ### Steps to install the SATA backplane:
 
 1. [Remove the top case](#removing-the-top-case).
-2. Unscrew and remove the 2.5" drive cage's cover.
+2. Unscrew and remove the cover of each 2.5" drive cage you're installing a SATA backplane into.
 
-![2.5" SATA drive cage cover](./img/25-drive-cover.webp)
+![2.5" SATA drive cage cover](./img/25-drive-covers.webp)
 
 3. Unscrew and remove the 2.5" drive cage.
```
leviport, line 253: Cage should be plural

src/models/nebula49-1/assembly.md
```diff
@@ -318,17 +314,17 @@ The 140mm bottom case fan (`BQ SIW4-14025-LF-PWM`) is preinstalled, but can be r
```
leviport, line 314: Step 4 above should be updated like nebula36 (no acrylic spacer, plus the new wording)

src/models/nebula49-1/assembly.md
```diff
@@ -318,17 +314,17 @@ The 140mm bottom case fan (`BQ SIW4-14025-LF-PWM`) is preinstalled, but can be r
 
 ## Installing side intake fans:
 
-nebula36 supports up to two 120mm side intake fans mounted to the GPU brace. One 120mm fan (`BQ SIW4-12025-MF-PWM`) is sold as an optional add-on at the time of purchase.
+nebula49 supports up to two 120mm side intake fans mounted to the GPU brace. One 120mm fan (`BQ SIW4-12025-MF-PWM`) is sold as an optional add-on at the time of purchase.
```
leviport, line 317: If two fans become an option, this sentence should be updated

src/models/nebula49-1/assembly.md
```diff
@@ -337,10 +333,11 @@ nebula36 supports up to two 120mm side intake fans mounted to the GPU brace. One
 ![Mounted side fan](./img/side-fan-mounted.webp)
 
 5. Plug the side fan into one of the headers on the GPU brace's fan splitter board. The cearby velcro strap can be used to hold the fan's cable.
```
leviport, line 335: s/cearby/nearby/g

leviport (changes requested):
Review complete, sorry it's scattered across multiple comments.

src/models/nebula49-1/assembly.md
jacobgkau: Since we sell the SATA backplanes individually, someone might be installing only one. I don't want to have to repeat "for each backplane you're installing" on every single line of this section, so I was trying to make it written in singular and then just say "do the section for each one." I'll try to make it more clear.

jacobgkau:
Addressed all the feedback.

Regarding the side intake fans, I had tried to make it vague since we didn't know what the final decision was going to be on that (missed the line in the GPU brace section that you pointed out.) However, it looks like we now offer two or zero as the only options, so I've updated the copy to reflect that.

leviport (approved):
Looks great!

PR #193, opened 2023-06-29
Update board options

ahoneybun:


jacobgkau (changes requested):
The screw hole sizes and motherboard shapes are compatible between DTX/Mini-DTX and ATX: https://en.wikipedia.org/wiki/Mini-ITX#/media/File:Comparison_ATX_%C2%B5ATX_DTX_ITX_mini-DTX.svg 

It doesn't really matter that the sales/marketing teams didn't care to add the DTX sizes to their materials, it makes sense for us to list them in the tech docs since they're compatible.

ahoneybun:
I'm good closing this in this case, thanks for the link!

PR #194, opened 2023-07-05
nebula19/36/49 (assembly): Rename GPU brace to side fan bracket

jacobgkau:
We don't intend to provide GPU brace fingers for Nebula, and people have gotten confused about why it's called a GPU brace if it doesn't directly brace the GPU vertically, so we're renaming it to something more reflective of its actual usage, which is apparently only to hold the side fans (which we recommend to be intake, but at least one person has already wanted to use for exhaust instead.)


leviport (approved):

jacobgkau:
Going to go ahead and get this merged to minimize further confusion.

PR #195, opened 2023-07-10
fix dimesions

thomas-zimmerman:
customer notices the dimensions were off


jacobgkau:
You just copy/pasted the format from the sales page, it doesn't match the other tech docs pages. I'll fix it.

jacobgkau (approved):

PR #197, opened 2023-07-18
Darter Pro 9 (darp9)

jacobgkau:
This adds the Darter Pro 9 (darp9) to tech docs.

The external overview is mostly reused from darp8 since the chassis are identical and marketing reused their port photos. The top-down shot is a new addition from the last generation, and uses the recently-updated shot with the new Super key.

**Battery replacement warning copy:**

Noteworthy is the last commit (64a621dce8cf8bc1618a268f57783a24b3006a79) updating the battery warning copy. The purpose of this update was to make it less of a wall of text and less nonlinear (the current copy having "This product contains a rechargeable battery." three quarters of the way through the paragraph reads awkwardly.)

The changes are mostly splitting it into multiple paragraphs, reordering some of the sentences between the paragraphs, and adding the battery part number (along with one more new warning about third-party replacements since we previously just said to contact System76 for replacements.)

This copy was originally provided by @bjornidentity as a condition for including the battery replacement instructions in lemp9's section, so I wanted to call special attention to the change in case we want to get additional approvals on it.


leviport:
I pushed a commit to change LCD toggle to mic mute since I had the layout PDF handy and could make the icon file easily.

jacobgkau:
Pushed an additional commit updating the RAM and SSD specs on the repairs page to match the spec overview, since I'd missed that before.

src/models/darp9/internal-overview.md
```diff
@@ -0,0 +1,28 @@
...
+## Bottom components and connectors:
+
+![Internal Components](./img/components-highlighted.webp)
+
+- RAM slots are highlighted in purple
+- Wireless card is highlighted in yellow
+- Fan connector is highlighted in red
+- Webcam/microphone connector is highlighted in orange
+- M.2 SSDs are highlighted in dark green
+    - Innermost slot supports PCIe NVMe Gen 3 and SATA III
```
leviport, line 12: They're both just PCIe gen4 now, right?

leviport:
I think this is good to go after the NVMe slot spec is updated.

leviport (approved):
:metal:

PR #200, opened 2023-08-07
nebula49-1: Remove EATX motherboard compatibility

jacobgkau:
It was discovered that EATX is not actually compatible (due to the top PCIe cutout not being present), so it was removed from the sales page and should be removed from tech docs as well. A `nebula49-2` with fixed EATX compatibility is planned.


jacobgkau:
Since this is being fixed anyway, it was decided it won't be waiting for another revision; existing customers will be contacted to offer a replacement once the update is complete. This spec update isn't needed anymore.

PR #201, opened 2023-08-10
Update memory specs

ahoneybun:
I have a customer that installed 5600Mhz RAM and it only goes up to 5200Mhz, the CPU also states that the max is 5200Mhz:

https://ark.intel.com/content/www/us/en/ark/products/232126/intel-core-i51340p-processor-12m-cache-up-to-4-60-ghz.html

### DIFF
--- a/src/models/darp9/README.md
+++ b/src/models/darp9/README.md
@@ -32,7 +32,7 @@ The System76 Darter Pro is a laptop with the following specifications:
         - 1x HDMI 2.1
         - 1x DisplayPort 1.4a over USB-C
 - Memory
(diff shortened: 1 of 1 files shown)


jacobgkau:
This sounded familiar, and we had to make a similar correction for serw13: https://github.com/system76/tech-docs/pull/184

jacobgkau (approved):
The sales page currently lists 4800 MHz, but a customer saw it operating at 5200 MHz, and we've got systems doing that speed after https://github.com/system76/firmware-open/pull/430.

The Clevo manual's 5600 MHz spec was evidently incorrect, similar to serw13 that was addressed in https://github.com/system76/tech-docs/pull/184.

PR #203, opened 2023-08-11
Add additional USB-C video out

ahoneybun:
This will need a new image of the right side noting that the USB-C port has DisplayPort support.


jacobgkau (approved):
Confirmed on the lab unit (3080 Ti GPU) that the side port supports video.

I don't think this requires a new photo. Before, the Thunderbolt port was just labeled as such, which worked because Thunderbolt includes video output capability. Now, both ports are listed as supporting DisplayPort, so there can be no confusion about whether one or the other includes support.

I also don't think we necessarily need to list the second port as not supporting charging based on no other sections having that note on more than one port, but I'm okay including it so it doesn't look like it might right next to the other entry that has the note.

PR #216, opened 2023-10-05
Change shroud to duct in all Thelio docs

ahoneybun:
The term shroud is only used in the tech docs and instead it is called the duct in production and inventory so changing this to match everywhere else.


thomas-zimmerman (approved):
This does move the docs to call the part what we use in production. Testing in mdbook, the images all work.

jacobgkau:
I updated the internal overview pages to match, since they still said shroud. The find & replace you did broke capitalization consistency on the legacy Major repairs page, so I tried to clean that up a bit as well.

jacobgkau (approved):
Reviewing internal communications, it seems like mechanical engineering also calls it a duct more often than a shroud. I'm fine with this change.

PR #220, opened 2023-11-09
Add model numbers for more batteries

leviport:


leviport:
I think that's everything. I also made everything more consistent, based on the ones that already had battery model numbers. The only oddballs are oryp9/10 and serw13 since the 15" and 17" had slightly different batteries with different model numbers.

Also threw in a minor correction on the addw1 tech specs (CPU said it was U-class, I corrected it to H-class).

It might be easier to rebase master into this after Spark is merged, so feel free to leave this until that is in.

jacobgkau:
The kudu6 section was one of the more recent ones, and I'm thinking having the model number as a sub-bullet looks better than having it in parenthesis.

image

I'm going to go ahead and make that the formatting for all of them, if you're okay with that.

leviport:
Totally, I think that looks better too. I actually set out doing it like that, but I found more instances of it in parentheses, so I just unified based off of what was already more prevalent.

jacobgkau:
Since this PR has a few unrelated fixes, it needs to be rebased and merged (not squashed).

It's probably best if someone else reviews this now since I've made significant changes.

XV-02 (approved):
Each change looks accurate, and the battery information appears correct on every system I checked against.

ahoneybun (approved):

PR #225, opened 2023-12-20
Meerkat 8 (meer8)

jacobgkau:
This PR will add the Meerkat 8 (meer8) to tech docs.


leviport (approved):
I noticed a lot of the overview photos are with the meer7 style lid, but I don't think that's a blocker. If new product photos with the slightly different lid are ever taken, we should change these out with those.

Everything else looks great!

PR #227, opened 2024-01-03
Swap backplane with board to match production and build

ahoneybun:
This PR swaps backplane for board to correct name of the component.


leviport:
I believe both terms are correct for that component. What makes you say that "board" is more correct than "backplane"?

ahoneybun:
> I believe both terms are correct for that component. What makes you say that "board" is more correct than "backplane"?

Production says backplane is incorrect and board is correct, like Thelio Io **board**.

leviport:
This README is obviously out of date, but the old Io board was said to "provide a SATA backplane" here: https://github.com/system76/thelio-io-hardware/blob/master/README.md?plain=1#L5

Since this PCB has the SATA functionality broken out into a standalone component, calling this a "SATA backplane" does not seem incorrect to me. Was there another reason for this change?

ahoneybun:
> This README is obviously out of date, but the old Io board was said to "provide a SATA backplane" here: https://github.com/system76/thelio-io-hardware/blob/master/README.md?plain=1#L5
> 
> Since this PCB has the SATA functionality broken out into a standalone component, calling this a "SATA backplane" does not seem incorrect to me. Was there another reason for this change?

The only reason is that both Production and Build call it SATA board and when backplane is mentioned they do not know what the part is. This is to get everyone on the same page about the component name.

jacobgkau (changes requested):
With respect to the production team, they don't get to unilaterally decide what the products they're building are called.

We are selling this part on the website as a "SATA Backplane":

image

ahoneybun:
Per @s-mcnees this naming comes from the vendor who makes the product and SATA board is used for ordering the product as well. If there are any other issues with this change please reach out to her.

jacobgkau:
> this naming comes from the vendor who makes the product

This is our product, we decide the name, not a vendor.

leviport:
"Board" is a generic word that describes a PCB, so referring to the component as a board does not mean it's the name of the product. A Lemur Pro is a Lemur Pro, but it is often referred to as a "laptop" or a "computer" or a "machine." Thelio Io is frequently referred to as an "Io board," but that's not its name.

I don't think this change is necessary, and I don't think "SATA backplane" was causing any confusion for readers.

leviport:
Per an internal discussion, the official name of the part is SATA Backplane, so I'm going to close this PR.

PR #229, opened 2024-01-13
Launch 3 (launch_3)

jacobgkau:
This adds the Launch 3 (launch_3) to tech docs.


jacobgkau:
@leviport I noticed we have the depth of launch_2 in the tech docs as 3.061cm, but the launch_3 sales page currently says 30.6mm. Do you know if there was actually a change in depth, or if we just decided to drop the last digit of specificity due to manufacturing tolerances not being tight enough to justify including it?

leviport:
There were no dimension changes. Rounding off a thousandth of a millimeter seems fine, but keeping it to match launch_2 also doesn't seem objectionable. I'm good either way, up to you.

leviport (approved):
Looks great! I didn't see any issues.

ahoneybun (approved):

PR #231, opened 2024-01-19
launch_heavy_3: Specs, overview, & repairs

jacobgkau:
This adds the Launch Heavy 3 (launch_heavy_3) to tech docs.

The most notable adjustment was that the vertical keycaps now have their switches rotated 90 degrees. That means the switches no longer come out with the keycaps when pulling those, but it also means special care has to be taken to orient the switches properly during reinstallation.

Also of note is that we skipped from Heavy 1 to Heavy 3. There was no Heavy 2; the version number was just incremented to match the regular Launch (and the primary Launch PCB).


src/models/launch_heavy_3/README.md
```diff
@@ -0,0 +1,47 @@
...
+- Layout
+  - Modified ANSI US QWERTY
+- Custom Configurations
+  - Customize layout and lighting in firmware with the [System76 Keyboard Configurator](https://github.com/pop-os/keyboard-configurator)
+  - [Available on Linux, Windows, and macOS](https://system76.com/accessories/launch/download)
+- Integrated Hub
+  - 2x USB 3.2 Gen 2 Type-C
+  - 2x USB 3.2 Gen 2 Type-A
+- Firmware
+  - [Open-source QMK-based firmware](https://github.com/system76/qmk_firmware/tree/master/keyboards/system76/launch_heavy_1/)
```
XV-02, line 42: I think we need to change this URl to the one for the Launch_Heavy_3: https://github.com/system76/qmk_firmware/tree/master/keyboards/system76/launch_heavy_3

src/models/launch_heavy_3/repairs.md
```diff
@@ -0,0 +1,169 @@
...
+3. Carefully line up the main PCB with the switch pins and place it in the chassis, starting with the USB-C port on the top edge.
+    - Evenly press the PCB down until all of the switch pins are fully seated in their sockets.
+    - It may help to pick up the keyboard and pinch to press both the PCB and the switches at the same time.
+    - If you feel excessive resistance, stop, remove the PCB, and ensure that all of the switch pins are lined up.
+    - The switch pins can be seen from the back of the sockets when they are fully seated, highlighted below.
+    
+![Switch pins visible through sockets](./img/pcb-switch-pins.webp)
+
+4. Repeat the previous step for the numpad PCB.
+5. Insert the ribbon cable into the white connector on both PCBs, then slide the black latches shut.
```
XV-02, line 168: I would add language here to stress that the user should ensure that the cable is clear of the screw-post.

XV-02 (commented):
This is looking stellar as usual, save for one incorrect link, and adding in language highlighting that there is a correct way to reinstall the ribbon cable for the ten-key.

src/models/launch_heavy_3/repairs.md
jacobgkau: I generally like to assume readers aren't stupid and don't need to have physically obvious things pointed out, but I agree we can specify the orientation of the ribbon since it can be installed either way, and I myself had to reference the pre-removal photo to remember the correct orientation when I was putting it back together last night.

XV-02: That's fair. I think I just look at it and... recognize that we have failed to develop the most intuitive design with that cable placement.

jacobgkau: I don't think the cable placement's too bad, we mechanically wanted a screw hole there and the ribbon gets out of the way for it.

Added a note.

XV-02 (approved):
Looks rock-solid.

PR #233, opened 2024-01-27
pang14: Specs, overviews, repairs

jacobgkau:
This adds the Pangolin 14 (pang14) to tech docs.

We've been adding an increasing number of part/model numbers to the descriptions at the start of each section on the repairs page. For this model, I've tried separating those numbers out into their own place in the description to make it more organized.


ahoneybun:
For some reason the main list is not updated but the side list is?

Screenshot from 2024-01-29 07-37-08

jacobgkau:
@ahoneybun The lists aren't connected in any way, I just forgot to update the homepage list. (It usually goes that way since the sidebar list needs to be updated for the pages to render at all.) Fixed now.

src/models/pang14/README.md
```diff
@@ -0,0 +1,57 @@
...
+- [External Overview](./external-overview.md)
+- [Internal Overview](./internal-overview.md)
+- [Parts & Repairs](./repairs.md)
+
+![Pangolin](./img/pang14.webp)
+
+The System76 Pangolin is a laptop with the following specifications:
+
+- CPU
+    - Supports AMD Ryzen (Rembrandt) Mobile CPUs
```
ahoneybun, line 12: I believe this is a different CPU generation from the pang12 and pang13?

ahoneybun: Looks like it should be Phoenix.

ahoneybun (dismissed):
Looks good I just think it is the CPU gen.

src/models/pang14/README.md
jacobgkau: Yes, good catch.

src/models/pang14/README.md
```diff
@@ -0,0 +1,57 @@
...
+- BIOS
+    - Winbond W25R256JW flash chip
+        - WSON-8 form factor
+    - Programmed with System76 firmware (non-open)
+- EC
+    - ITE IT5571
+        - Programmed with non-open EC firmware
+- Graphics
+    - GPU: AMD Radeon 780M
+    - eDP display: 16.1" 1920x1080@144Hz LCD
```
leviport, line 23: New 16" panel is only 60Hz

leviport (changes requested):
One correction, but otherwise looks great! I like the addition of the part numbers in the repairs page.

src/models/pang14/README.md
jacobgkau: Of course. Fixed.

leviport (approved):
:+1:

PR #242, opened 2024-04-09
Lemur Pro 13 (lemp13)

jacobgkau:
This adds the Lemur Pro 13 (lemp13) to tech docs.


leviport (changes requested):
Two things, but otherwise looks excellent!

jacobgkau:
The lemp13's photos with the bottom sticker visible (the speaker photo and the bottom panel screws photo) have been replaced after reshooting those photos with a System76 product sticker applied, now that we got one of those in-house.

leviport (approved):
:metal:

jacobgkau:
I'm merging so I can start the next section's branch, but if support finds any issues, we can address them there.

PR #249, opened 2024-05-22
lemp13 (external overview): Add placeholder photo of new keyboard

jacobgkau:
This photo's just a placeholder to show the currently shipping keyboard in the lemp13-b. We'll swap it out with the marketing overhead photo after their photoshoot on the 29th.

Not sure if we need/want to update the sidebar & model page heading to reflect that the section is for both lemp13 and lemp13-b. The firmware is different, but we didn't update the stickers, so someone looking at a lemp13-b will just see "lemp13" written on it. I don't think there will be any confusion about whether this section applies to them or not (it's not like there are any other lemp13 sections to mix it up with).


ahoneybun (commented):
Do we want to include a command to tell them if they are on `lemp13` or `lemp13-b`?

leviport (approved):

jacobgkau:
> Do we want to include a command to tell them if they are on `lemp13` or `lemp13-b`?

I don't see why that would be necessary in the context of this photo. They can see which one they're on based on the keyboard layout, which we're providing photos of.

If we decide it's necessary to provide more information about the firmware difference later, we can potentially mention a command.

PR #253, opened 2024-06-06
darp10 (external overview): Flip labels for Thunderbolt & USB-C

### LINKED ISSUE #252: darp10-b labeling of USB-c wrong?
it looks like thunderbolt and usb-3.2 labeling of the right side of the darp10-b is switched, if you believe the labeling on the case I have in front of me (which matches the picture in the document here). That's in external-overview.md

jacobgkau:
Fixes #252.


leviport (approved):
I definitely missed that in my review. Thanks for the quick fix!

PR #256, opened 2024-06-24
Correct the Amp for the charger

ahoneybun:
PXL_20240624_142414960

PXL_20240624_142421046


leviport (approved):
330/19.5 = 16.92A, so I believe this was a typo. Thanks for the correction

jacobgkau (approved):

PR #257, opened 2024-06-29
Thelio Mira B4-N3 (thelio-mira-b4-n3)

jacobgkau:
This adds the Thelio Mira B4-N3 (thelio-mira-b4-n3) to tech docs.


jacobgkau:
I still need to update the GPU length clearance once I get those numbers. I also need to double-check the following items:

- CPU fan model number (both front & back; I currently have them down as the same based on the CPU cooler manual)
- Ethernet controller model number

Everything else should be ready to review.

jacobgkau:
Updated the GPU clearance & confirmed the other two items were already correct.

jacobgkau:
Fixed the link on the homepage not working.

src/models/thelio-mira-b4-n3/repairs.md
```diff
@@ -0,0 +1,613 @@
...
+3. Unscrew the four screws holding the drive cage onto the chassis (highlighted red above), then remove the drive cage from the chassis.
+4. Unscrew the four SATA backplane screws from the drive cage, then push the SATA backplane out through the front of the drive cage.
+
+![SATA backplane screws](./img/sata-backplane-screws.webp)
+
+5. Place the new SATA backplane into the drive cage and secure it with the four SATA backplane screws, then reinstall the drive cage (if necessary).
+6. Plug the 4-pin Berg power connector and SATA data cables into the SATA backplane.
+    - The order of the SATA data cables shouldn't matter as long as your operating system and software is configured to address disks by UUID (e.g. `/dev/disk/by-id/...`) instead of by letter (e.g. `/dev/sd_`).
+    - By default, the `DATA1` port connects to SATA port 4 on the motherboard, and the `DATA0` port connects to SATA port 5.
+    - Ensure the cables are not plugged into SATA ports 2 and 3 (the bottom two ports), as these ports are unavailbable when the M.2 slots are fully populated.
```
ahoneybun, line 532: From the board it looks like the bottom M.2 is what turns those ports off **but** if you use an M.2 SATA and not an NVMe (NVMe for the other 3) those SATA ports will still work though. 

screenshot-2024-07-08-18-44-29

ahoneybun (dismissed):
Just a little note and perhaps something to think about.

src/models/thelio-mira-b4-n3/repairs.md
jacobgkau: It's irrelevant as long as someone doesn't plug into the bottom two ports, which never needs to happen because we only have two 2.5" slots. I'm not sure it's worth complicating the wording just to explain a situation that physically can't/doesn't need to happen in the chassis.

jacobgkau: Updated to match line 281 even though explaining this is completely pointless.

src/models/thelio-mira-b4-n3/repairs.md
```diff
@@ -0,0 +1,613 @@
...
+    - The replacement PSU should be installed with the fan facing the bottom of the case.
+    - Place the PSU on the rubber post and hold it against the screwholes in the back of the case while attaching it.
+7. Connect the power cables to the new power supply and GPU (if installed), then reinstall the PSU brace, side brace, and top case.
+
+## Replacing the Thelio Io board:
+
+Named after Jupiter's moon Io, the Thelio Io daughterboard handles the front power button and fan control. If the Thelio Io board becomes defective, it can be replaced using the instructions below.
+
+**Part numbers:**
+- Thelio Mira B4-N3 uses [Thelio Io version 2](https://github.com/system76/thelio-io/tree/92c2a60b6d6acc0f9d51d819cba07ad28e2b03f3) ([PCB revision thelio_io_2.3](https://github.com/system76/thelio-io-hardware/tree/91814370002f762ac01cd1dd99dcbb43336352ab/pcb-thelio-io).
```
leviport, line 456: Looks like this line is missing a closing parentheses `)` for the PCB revision note.

src/models/thelio-mira-b4-n3/repairs.md
```diff
@@ -0,0 +1,613 @@
...
+    - `FANOUT1`/`CPU FAN` - to the top fan splitter (connecting to the [CPU fans](#replacing-the-cpu-fans)).
+    - `FANOUT4`/`AUX FAN` - to the bottom fan splitter (below the [CPU fan splitter](#replacing-the-cpu-fans)).
+    - `FANOUT2`/`INTAKE FAN` - to the [bottom case fan](#replacing-the-bottom-case-fan).
+    - `FANOUT3`/`GPU FAN` - to the [side brace](#removing-the-side-brace) fan splitter.
+        - May be disconnected by default if no side fan is ordered as factory configuration.
+    - `FANIN1`/`MOTHERBOARD FAN INPUT` - to the `CPU_FAN` header at the top right of the motherboard, highlighted cyan below.
+    
+    ![Motherboard CPU output header](./img/motherboard-fan-header.webp)
+    - `POWER1`/`POWER INPUT` - to the [power supply](#replacing-the-power-supply).
+        - The white plastic backing needs to be held away from the connector to unplug this cable from the Thelio Io board.
```
leviport, line 491: "white plastic backing" seems like an awkward descriptor. Would "white plastic locking tab" or "white plastic latch" work better?

leviport (changes requested):
A couple small things, but otherwise looks great!

src/models/thelio-mira-b4-n3/repairs.md
jacobgkau: I think "white plastic clip" is fine, and probably more natural than "backing."

src/models/thelio-mira-b4-n3/repairs.md
ahoneybun: Thank you and I do think it has a point if someone wants to add a PCIe device that uses SATA for power and will find those ports disabled but if they use an M.2 SATA instead of M.2 NVMe they will remain active.

jacobgkau:
I've added a second commit to fix the Realtek RTL8125 link in the other sections that linked to it. **Please use Rebase & Merge when merging this PR.**

leviport (approved):
:metal:

PR #258, opened 2024-07-09
Add note about Fn+F4 for the 14" model and 16" model

ahoneybun:
The darp10-b has the backlight set for Fn+F4 but we only note about the darp10 which has it on Fn+* due to the Numpad.

### DIFF
--- a/src/models/darp10/external-overview.md
+++ b/src/models/darp10/external-overview.md
@@ -75,22 +75,23 @@
 
 ## Keyboard shortcuts:
 
-|Key                        |Shortcut|Action                             |
-|---------------------------|--------|-----------------------------------|
-|![Fn-F1](./img/fn-f1.png)  |Fn+F1   |Toggle trackpad                    |
-|![Fn-F2](./img/fn-f2.png)  |Fn+F2   |Mic mute                           |
-|![Fn-F3](./img/fn-f3.png)  |Fn+F3   |Mute                               |
-|![Fn-F5](./img/fn-f5.png)  |Fn+F5   |Volume down                        |
-|![Fn-F6](./img/fn-f6.png)  |Fn+F6   |Volume up                          |
-|![Fn-F7](./img/fn-f7.png)  |Fn+F7   |Toggle displays                    |
-|![Fn-F8](./img/fn-f8.png)  |Fn+F8   |Screen brightness down             |
-|![Fn-F9](./img/fn-f9.png)  |Fn+F9   |Screen brightness up               |
-|![Fn-F10](./img/fn-f10.png)|Fn+F10  |Toggle webcam                      |
-|![Fn-F11](./img/fn-f11.png)|Fn+F11  |Toggle airplane mode               |
-|![Fn-F12](./img/fn-f12.png)|Fn+F12  |Suspend                            |
-|![Fn-F12](./img/fn-dia.jpg)|Fn+`    |Play/Pause                         |
-|![Fn-*](./img/fn-star.png) |Fn+*    |Toggle keyboard backlight          |
-|![Fn-/](./img/fn-slash.png)|Fn+/    |Cycle keyboard brightness levels   |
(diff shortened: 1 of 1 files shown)


src/models/darp10/external-overview.md
```diff
@@ -80,6 +80,7 @@
 |![Fn-F1](./img/fn-f1.png)  |Fn+F1   |Toggle trackpad                    |
 |![Fn-F2](./img/fn-f2.png)  |Fn+F2   |Mic mute                           |
 |![Fn-F3](./img/fn-f3.png)  |Fn+F3   |Mute                               |
+|![Fn-F4](./img/fn-star.png)  |Fn+F4   |Toggle keyboard backlight          |
```
leviport, line 83: This should probably also have an asterisk and a footnote, since the 16" does not have a toggle hotkey on F4

leviport (changes requested):

src/models/darp10/external-overview.md
```diff
@@ -89,8 +90,11 @@
 |![Fn-F11](./img/fn-f11.png)|Fn+F11  |Toggle airplane mode               |
 |![Fn-F12](./img/fn-f12.png)|Fn+F12  |Suspend                            |
 |![Fn-F12](./img/fn-dia.jpg)|Fn+`    |Play/Pause                         |
-|![Fn-*](./img/fn-star.png) |Fn+*    |Toggle keyboard backlight          |
+|![Fn-*](./img/fn-star.png) |Fn+*    |Toggle keyboard backlight*         |
```
leviport, line 93: Actually, the next three hotkeys probably all need the note for being on the 16" model only.

Also, you'll need a different marker for the 14" and 16" notes. Otherwise I can't tell which asterisk corresponds to which note. Two asterisks should suffice on one of the model.

leviport (changes requested):

ahoneybun:
I think that is the best I can get with Markdown.

src/models/darp10/external-overview.md
```diff
@@ -89,8 +90,11 @@
...
-|![Fn--](./img/fn-minus.png)|Fn+-    |Decrease keyboard brightness       |
-|![Fn-+](./img/fn-plus.png) |Fn++    |Increase keyboard brightness       |
+|![Fn-*](./img/fn-star.png) |Fn+*    |Toggle keyboard backlight**        |
+|![Fn-/](./img/fn-slash.png)|Fn+/    |Cycle keyboard brightness levels** |
+|![Fn--](./img/fn-minus.png)|Fn+-    |Decrease keyboard brightness**     |
+|![Fn-+](./img/fn-plus.png) |Fn++    |Increase keyboard brightness**     |
 |1                          |Fn+1    |Toggle fan between max/automatic   |
+
+- For the 14" version of the Darter Pro (darp10-b) *
+- For the 16" version of the Darter Pro (darp10) **
```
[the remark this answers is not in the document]
ahoneybun: That looks like it works! I was a bit wary to add HTML in there but I guess it works.

leviport (approved):
This looks good to me, but let's make sure we get Jacob's blessing before we merge, just in case he prefers this be done a different way.

jacobgkau:
This is what you had:

screenshot-2024-07-11-22-28-42
screenshot-2024-07-11-22-28-47

Not optimal that the top asterisk and the meaning at the bottom can't be in view at the same time when maximized on a normal 1080p display. I'm updating it to look like this instead:

screenshot-2024-07-11-22-32-21
screenshot-2024-07-11-22-32-26

No scrolling, and matches how it's phrased on the rest of the page better. We could get away with just "14:" and "16:", but I think we might as well do the whole thing since the icons are taking up the vertical space anyway.

leviport:
I like that better as well!

leviport (approved):

jacobgkau (approved):

leviport (approved):

PR #260, opened 2024-08-30
Add BE200 Wifi card option to tech-specs for laptops that offer it.

leviport:
I copied the formatting from the GPU section on laptops with multiple GPU options. Feel free to make changes if there's a better way to present this option.


thomas-zimmerman (approved):
LGTM for now; do we know what kernel version support is good for on the BE200?

jacobgkau (approved):
We're a little inconsistent, but we generally list options in order from most premium to least premium. I've swapped the WiFi cards around, and fixed up a couple other components on the serw13 page to follow that rule.

Other than that, looks good to me.

PR #261, opened 2024-09-03
Remove note about 2nd fan as it also spins 

ahoneybun:
This matches the pang13/14


leviport (approved):
Good call, I forgot about this note

PR #269, opened 2024-10-15
bonw15 (specs): Update with bonw15-b notes

### LINKED ISSUE #268: bonw15 docs update: bonw15-b
The Bonobo WS was updated to bonw15-b, which includes a TB5 controller. We aren't yet shipping bonw15-b with Open Firmware, since there are still some bugs to workout with this new TB controller. I don't think making an entire new model for bonw15-b is sensible, but we will need to mention the new TB5 controller and port capabilities, as well as a mention that these shipped with proprietary firmware & EC.

Once we have the TB5 controller issues worked out and push the Open Firmware transition, we'll probably need to make another update to reflect this.

jacobgkau:
Closes #268.

### DIFF
--- a/src/models/bonw15/README.md
+++ b/src/models/bonw15/README.md
@@ -1,4 +1,4 @@
-# Bonobo WS (bonw15)
+# Bonobo WS (bonw15 & bonw15-b)
 
 - [External Overview](./external-overview.md)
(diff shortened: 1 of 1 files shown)


leviport (approved):
Looks great!

PR #271, opened 2024-10-30
thelio-r3-n1: Update maximum memory capacity & frequency

Integral-Tech:
- Reference: https://www.msi.com/Motherboard/MPG-B650I-EDGE-WIFI/Specification

> Memory: 2x DDR5, Maximum Memory Capacity 128GB
Memory Support DDR5 7200+(OC)/ 7000(OC)/ 6800(OC)/ 6600(OC)/ 6400(OC)/ 6200(OC)/ 6000(OC)/ 5800(OC)/ 5600(OC)/ 5400(OC)/ 5200(OC)/ 5000(OC)/ 4800(JEDEC) MHz

### DIFF
--- a/src/models/thelio-r3-n1/README.md
+++ b/src/models/thelio-r3-n1/README.md
@@ -38,7 +38,7 @@ The System76 Thelio is a desktop with the following specifications:
 - Expansion
     - 1x PCIe 4.0 x16 (GPU slot)
 - Memory
-    - Up to 64GB (2x32GB) dual-channel DDR5 DIMMs @ 4800 MHz
+    - Up to 128GB (2x64GB) dual-channel DDR5 DIMMs @ 4800 MHz (JEDEC) or 7200+ MHz (OC)
     - Tested with the following RAM modules (may ship with other tested modules):
         - [Crucial CT32G48C40U5](https://www.crucial.com/memory/ddr5/ct32g48c40u5#spec) (32GB/stick)
 - Networking


jacobgkau:
> Memory Support DDR5 7200+(OC)/ 7000(OC)/ 6800(OC)/ 6600(OC)/ 6400(OC)/ 6200(OC)/ 6000(OC)/ 5800(OC)/ 5600(OC)/ 5400(OC)/ 5200(OC)/ 5000(OC)/ 4800(JEDEC) MHz

I see in that line that 4800 is the only speed listed as JEDEC (which means native speed). All the higher speeds require overclocking. I guess we'll need to test if overclocking works and is stable on a Thelio R3-- it does appear from internal conversations like we might have went with 4800 partially for pricing reasons, so it might work fine. Not sure if any default BIOS settings would need to be changed to enable overclocking (XMP, etc).

I'm not sure why we weren't offering the system with 128GB as a config option if it supported that. It looks like we're also offering less than the motherboard maximum for R4, but its motherboard vendor only says it supports up to 96GB. It could just be a product class thing.

jacobgkau (approved):
I don't think we're going to be testing this at this point (the model's 2 versions old). I found indications that stability is hit-or-miss at higher capacities and speeds than what we shipped (e.g. [here](https://www.reddit.com/r/Amd/comments/130dame/any_of_you_used_the_msi_mpg_b650i_edge_wifi_what/)).

Updated the wording to clarify what was tested and what the motherboard theoretically supports.

Issue #275, opened 2025-01-02
Missing Dimensions from Astra page


ahoneybun:
Other models like the Thelio (Prime) have this on the bottom of the README page but Astra is missing it.

![screenshot-2025-01-02-12-44-19](https://github.com/user-attachments/assets/27751850-d722-4dff-ad63-cf73cea2fd29)

https://tech-docs.system76.com/models/thelio-astra-a1-n1/README.html

ahoneybun:
This is fixed in PR: https://github.com/system76/tech-docs/pull/276

PR #276, opened 2025-01-02
copy specs from Mira as it is the same chassis

ahoneybun:
I talked to production and the specs are the same for Astra, Mira, and Spark. 

This fixes issue https://github.com/system76/tech-docs/issues/275


jacobgkau (approved):
We don't even need to rely on copying, the Astra tech specs on https://system76.com/autonomous-vehicles include the dimensions. We just forgot to put them on the page. The numbers being added are correct.

PR #279, opened 2025-01-06
Added thelio-major-r5-n3 to current models.

jklgrasso:
The thelio-major-r3 tech docs are in both the current and previous models. I removed it from current and added the thelio-major-r5-n3 tech docs to the current models. Please double check that I did this right. :smile:


leviport (changes requested):
Please move the major-r3 links down to the Previous Models section, rather than just replacing it. We want major-r3 to still be discoverable.

ahoneybun:
@leviport it looks like the major-r3 is currently in current and previous sections so this PR looks good.

ahoneybun:
screenshot-2025-01-06-20-50-55

leviport:
Ah, I missed it in Previous. This is good then. However, you should also update src/README.md so that the ToC on the first page matches the index on the side bar.

leviport (approved):

jacobgkau:
Thank you for the fix, it looks like this was initially done properly when R5-N3 was added but had been rolled back in one of the Astra commits: https://github.com/system76/tech-docs/commit/6077873c138414565fa57a99ec9526fa2fc36d34

PR #283, opened 2025-01-23
add system76-driver-nvidia package to the steps

ahoneybun:
That was missing so the system would try to use nouveau instead.

### DIFF
--- a/src/models/thelio-astra-a1-n1/ubuntu-24.04.md
+++ b/src/models/thelio-astra-a1-n1/ubuntu-24.04.md
@@ -43,7 +43,7 @@ Thelio Astra uses an Ampere processor supporting AArch64 (ARM64) versions of Lin
     ```
 3. Install the System76 driver with the following command:
     ```
-    sudo apt install system76-driver
+    sudo apt install system76-driver system76-driver-nvidia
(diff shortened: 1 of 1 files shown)


jackpot51 (approved):

leviport:
`system76-driver` is a dependency of `system76-driver-nvidia`, so I'm not sure that it makes sense to have both next to each other. In https://support.system76.com/articles/system76-driver, we split `system76-driver-nvidia` out under a "If you have an NVIDIA gpu..." section, so maybe matching that here would be best?

ahoneybun:
@leviport I mean you could just install `system76-driver-nvidia` rather then both sure but since we only ship NVIDIA right now (and it won't hurt anything if you used AMD instead) I think this change is fine.

leviport:
Please don't merge tech-docs PRs without a QA review...

jacobgkau:
@ahoneybun Why did you request a review specifically from me, receive feedback from another member of QA, then merge this without either of us approving?

Only shipping NVIDIA and not hurting anything if you have AMD has nothing to do with one package being needed instead of both. If anything, it only reinforces the feedback you were given.

PR #284, opened 2025-01-23
thelio-astra-a1-n1 (Ubuntu 24.04 setup): Correct NVIDIA instructions

jacobgkau:
Fixes https://github.com/system76/tech-docs/pull/283, which was merged without full approvals.


leviport (approved):
Thank you for fixing that

thomas-zimmerman (approved):
I think that works well to be clear on the package to have installed.

PR #285, opened 2025-01-24
Add nix-shell file for mdbook

ahoneybun:
This matches up with the docs repo where with Nix installed you can just enter `nix-shell` and have `mdbook` ready to go.


jacobgkau (approved):
I still had to run `mdbook serve` inside of the Nix shell in order to start the web server, but this does appear to grab mdbook from the Nix repo instead of having to download it manually. (Note that since we update mdbook manually, the version may not always match with the version available from the Nix repo.)

jacobgkau:
If you want to merge this as-is, please make sure to use the `Rebase and merge` option so your extra merge commit doesn't show up.

ahoneybun:
That is expected and how the nix-shell works in the docs repo as well, you have to run `npm ci` and `npm start` to run the server locally. I can make the service start if we wanted to though.

jacobgkau:
I'm fine either way. Since mdbook is a single-file binary, using nix-shell to start it turns one command into two for me. But it sounds like it's doing what it's meant to be doing.

jacobgkau:
I'll just merge this as-is for now.

PR #286, opened 2025-01-25
specify to press OUT on ram tabs

kgilbert78:


jacobgkau:
If we just change "press" to "pull" like we currently have in the [Gazelle section](https://tech-docs.system76.com/models/gaze18/repairs.html#replacing-the-ram), would that solve the confusion for you?

"Press out" doesn't make a lot of sense on its own, and "out" still doesn't specify which direction (as it could be interpreted to mean "out from the motherboard," which would also be wrong). "Away from the RAM sticks" would be the most accurate, although it's verbose.

jacobgkau:
"Lift" would imply pulling them away from the motherboard, so I don't think we should use that word.

jacobgkau:
I've updated the wording to match the latest section (bonw16) and other recent sections.

jacobgkau (approved):

leviport (approved):

PR #289, opened 2025-01-31
add new CPU fan being used

ahoneybun:
This adds the CPU cooler that is being used in recent nebula19 cases as the optional CPU cooler.

I have not found a part number itself for the cooler or for the fans like we have it for the Noctua that was originally used. I added similar to have the have LCD panel options or Wi-Fi options such as in the latest Lemur Pro (lemp13).

### DIFF
--- a/src/models/nebula19-1/README.md
+++ b/src/models/nebula19-1/README.md
@@ -18,8 +18,11 @@ The System76 nebula19 is a desktop chassis (for DIY builds) with the following s
 - Cooling capacity
     - 2x 92mm CPU fans
         - Rear exhaust fan: Be Quiet! `PUW2-9225-MR-PWM`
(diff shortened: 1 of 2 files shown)


jacobgkau:
The fan model's right on the product page of the cooler. I'll add it.

Screenshot_2025-01-31_13-31-46

Will also add the date that we switched if I can find it.

jacobgkau (changes requested):
As I communicated in Slack, this change technically doesn't apply to nebula19-1. We had already bumped to -2 or -3 by the time this occurred.

If we were going to update this section, we'd also need to note it on the Repairs page, not just the overview. But it would be better if we could make an updated nebula19 section for the current version.

leviport (approved):

PR #293, opened 2025-02-20
thelio-astra-a1-n1: Update info about 1.1 revision

### LINKED ISSUE #292: thelio-astra-a1.1
We will soon be offering Astras with the other version of the motherboard (with 10Gb Ethernet instead of SFP+).  This motherboard variant is already present in Astra's tech-docs, but we'll need to lay it out by version numbers instead of date ranges. The versions are:
- thelio-astra-a1: uses ALTRAD8UD2-1L2Q with SFP+
- thelio-astra-a1.1: uses ALTRAD8UD-1L2T with 10Gb Ethernet

jacobgkau:
This was a fairly simple change to make since we already referred to the variants by motherboard model instead of by date in most locations.

Closes #292.


leviport (approved):
Looks good, thanks!

PR #297, opened 2025-03-25
Update ubuntu-24.04.md

thomas-zimmerman:
Add how to disable suspend if you have long running tasks on the system.


jacobgkau (commented):
Isn't this already done by the System76 Driver when it's installed a few lines before? (Because the Astra motherboard doesn't support suspend.)

https://github.com/pop-os/system76-driver/blob/3caaea1a28711827c4fcd37c53655c2630143307/system76driver/products.py#L1153

leviport:
Doesn't system76-driver already mask suspend? https://github.com/pop-os/system76-driver/pull/310

thomas-zimmerman:
If it is, it's not working. Demo system that Joe speed is working on had to have the services masked for their workload to stop breaking.

jacobgkau:
Re-open if necessary, but our testing found that the System76 Driver app is doing what it needs to for this, and suggested our partners may be doing something custom and not quite following the instructions here to install the driver.

PR #306, opened 2025-07-03
bonw16: Overview & repairs

jacobgkau:
This adds port diagrams, component/connector overviews, and a repair manual for the Bonobo WS 16 (bonw16).

Also made a small adjustment to wording in the bonw15 section (it has three ribbon cables for its keyboard, so the word "both" shouldn't have been used to refer to them).

### DIFF
--- a/src/models/bonw16/repairs.md
+++ b/src/models/bonw16/repairs.md
@@ -1,3 +1,283 @@
 # Bonobo WS (Parts & Repairs)
 
-A service manual for the Bonobo WS 16 (bonw16) is not yet available. Please reference the service manual for the previous version, the [Bonobo WS 15 (bonw15)](/models/bonw15/repairs.md).
+Many components in your Bonobo WS can be upgraded or replaced as necessary. Follow these step-by-step guides for instructions:
+
+- [Removing the bottom cover](#removing-the-bottom-cover)
+- [Replacing the battery](#replacing-the-battery)
+- [Replacing the RAM](#replacing-the-ram)
+- [Replacing an M.2/NVMe SSD](#replacing-an-m2nvme-ssd)
+- [Replacing the WiFi/Bluetooth module](#replacing-the-wireless-card)
+- [Replacing the CMOS battery](#replacing-the-cmos-battery)
+- [Replacing the speakers](#replacing-the-speakers)
+- [Replacing the fans/heatsink/thermal paste](#replacing-the-cooling-system)
+- [Replacing the keyboard](#replacing-the-keyboard)
+
+## Removing the bottom cover:
+
+Removing the cover is required to access the internal components. Prior to removing the cover, ensure the AC power is unplugged and all peripherals (such as USB drives or Kensington locks) are unplugged or removed from the system.
+
+**Part numbers:**
+- Bottom panel: `6-39-X5803-013`
+
+**Tools required:** Cross-head (Phillips) screwdriver  
+**Time estimate:** 5 minutes  
+**Difficulty:** Easy <span style="color:green;">●</span>
+
+### Steps to remove the bottom cover:
+
+1. Place the machine lid-side down.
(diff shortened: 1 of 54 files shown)


src/models/bonw16/repairs.md
```diff
@@ -1,3 +1,283 @@
...
+![Battery screws](./img/battery.webp)
+
+3. Unplug the white connector connecting the battery to the motherboard.
+4. Remove the battery, starting from the top edge with screw holes.
+5. When putting in the new battery, start with the plastic tabs along the bottom edge (opposite from the screw holes.)
+6. When plugging in the new battery, the red wire on the connector goes on the left, and the black wire goes on the right.
+
+## Replacing the RAM:
+
+The Bonobo WS 16 supports up to 192GB (4x48GB) of DDR5 SO-DIMMs running in dual-channel configuration at 5600MHz. The supported configurations are two or four SO-DIMMs (one or three SO-DIMMs are not supported configurations). If you've purchased new RAM, need to replace your RAM, or are reseating your RAM, follow these steps.
```
leviport, line 63: I think single-SODIMM configs are supported, but only in certain slots. Would it be helpful to include a slot order for RAM installation? It's not so bad when there are only two slots, but four gets a little more complicated. Here's the one from the manual, for reference:
Screenshot_2025-07-07_09-29-19

Also, I'm not sure the speed goes up to 5600MHz on this machine? If not, this also should be updated in README.md

src/models/bonw16/repairs.md
```diff
@@ -1,3 +1,283 @@
...
+3. To clear the CMOS, open the lid of the machine and hold down the power button for at least 15 seconds to discharge any residual energy in the system.
+4. Re-connect the CMOS battery and the main battery.
+5. Power up the Bonobo WS. The system may power itself off and on after initial boot; this is normal behavior when the CMOS has been reset.
+
+## Replacing the speakers:
+
+The system has three bottom-firing speakers, which can be removed and replaced individually.
+
+**Part numbers:**
+- Subwoofer:
```
leviport, line 161: Should this line be removed until we have the part number?

src/models/bonw16/repairs.md
```diff
@@ -1,3 +1,283 @@
...
+    - Screws #1-#12 are held captive, and will not completely detatch from the heatsink/fans.
+    - Screws #13 and #14 are slightly smaller than the rest of the screws, and are not held captive.
+    - Do not remove the smaller screws holding the fan covers onto the fans.
+
+![Thermal screws](./img/thermal-system.webp)
+
+4. Remove any clear tape covering the fan wires.
+5. Unplug the black fan connectors from the motherboard's fan headers, `J_VGAFAN1` (highlighted red above) and `J_CPUFAN1` (highlighted green).
+6. Remove the heatsink/fans from the case, being careful not to bend the heatsink pipes. It may take some pressure to break the seal of the thermal paste.
+7. Using a paper towel, remove the existing thermal paste from the CPU, GPU, VRAM chips, and heatsink. You may also use a small amount of rubbing alcohol if the old paste is dried or difficult to remove.
```
leviport, line 225: I believe what's used on the VRAM chips is a more delicate and less dense thermal pad, not liquid thermal paste like what is used on CPU/GPU dies directly. Are we sure we want to direct readers to remove these pads and use liquid paste on the VRAM chip packages? I imagine it'll work, but it seems like it could make quite a mess.

leviport (changes requested):
Just a couple things, but this otherwise looks great

src/models/bonw16/repairs.md
jacobgkau: All sections have been updated with new verbiage to explain the thermal putty. The bonw16 photos have been retaken to show the putty collection process.

leviport (approved):
The putty addition looks great!

PR #308, opened 2025-07-15
darp11 (external overview): Update captions on lid photo

jacobgkau:
This updates the captions on the darp11 lid photo, which was reused from the darp10 section but didn't have its captions updated in https://github.com/system76/tech-docs/pull/300.


leviport (approved):
Yep, I totally missed that on my first review. Thanks for catching and fixing it

PR #314, opened 2025-09-02
pang15 (specs): Correct dimensions

jacobgkau:
Levi took the following measurements in the lab:

IMG_20250902_145848_536

IMG_20250902_150000

Assuming it was typo'd when it was input into the sales page where I pulled the numbers from.


jacobgkau:
I'll merge this once we have confirmation on the [sales page](https://system76.com/laptops/pang15/configure) being updated with the same numbers, to avoid getting out of sync if there's further discussion.

Issue #302, opened 2025-06-21
GPL license isn't ideal for documentation


oxguy3:
I was interested in potentially reusing some of the content from this repository – I think the photos of System76 laptops and internal components could be really useful on, for example, Wikipedia and Wikimedia Commons. However, I noticed that the repository is licensed under the GNU Public License (GPLv3).

While the GPL is great for software, it is less ideal for written material and multimedia. For example, the GPL requires that the source code be made available, and defines source code as "the preferred form of the work for making modifications to it". Besides a few lines of CSS and a few config files, this repository doesn't contain anything that we would generally think of as "source code", so what does this requirement mean in this case? If I make a print copy of the manual, do I have to provide readers with the Markdown files? If I edit one of the photos, do I have to provide a .xcf or .psd file? GPL also requires that the entire text of the license be republished, which may be burdensome for, say, someone wanting to use a single photo in an article.

For these reasons, Wikimedia Commons [discourages uploading media with only a GPL license](https://commons.wikimedia.org/wiki/Commons:Copyright_tags/GNU_licenses), and instead prefers Creative Commons licenses. I think [CC BY-SA 4.0](https://creativecommons.org/licenses/by-sa/4.0/) would be a good fit. It requires attribution, but it allows the license to be linked or mentioned by name instead of requiring the entire text be included. It requires that any republications/modifications be published under the same license, but doesn't require source code to be shared.

So, would System76 be willing to change the license to CC BY-SA 4.0 (or a similar license)? Thanks!

jacobgkau:
I think CC BY-SA would be a good choice for the multimedia in the repo.

Just to address some of your hypotheticals:

> If I make a print copy of the manual, do I have to provide readers with the Markdown files?

No, you'd only have to include the URL of the GitHub repo, as per 6.b.2 of the GPL ("a written offer... to give anyone who possesses the object code... access to copy the Corresponding Source from a network server at no charge").

> If I edit one of the photos, do I have to provide a .xcf or .psd file?

I don't even provide .xcf or .svg files within the repository in the first place, so I'd personally consider the rendered images to be the "source" in the context of the website's code.

[...]

PR #315, opened 2025-09-03
Switch license from GPLv3 to CC BY-SA 4.0

jacobgkau:
Closes https://github.com/system76/tech-docs/issues/302.

This makes more sense for documentation, and also matches the license of https://github.com/system76/docs.

CC BY-SA 4.0 is [one-way compatible](https://wiki.creativecommons.org/wiki/ShareAlike_compatibility:_GPLv3) with the GPLv3, so it can still be used as GPLv3 if someone wants to do that.

98.2% of the commits and 99.9% of the lines of text in this repository have come from employees. All commits of actual content by non-employees have been corrections to statements of fact (not necessarily copyrightable); the only other commit by a non-employee was the author of mdBook removing a deprecated field from a CI document, which is trivial. Based on this, I believe we're legally in the clear to make this change. As mentioned above, we would be able to convert back to GPLv3 in the future if desired without any legal concerns due to the one-way compatibility.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,674 +1,427 @@
-                    GNU GENERAL PUBLIC LICENSE
-                       Version 3, 29 June 2007
-
- Copyright (C) 2007 Free Software Foundation, Inc. <https://fsf.org/>
- Everyone is permitted to copy and distribute verbatim copies
- of this license document, but changing it is not allowed.
-
-                            Preamble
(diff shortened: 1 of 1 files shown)


thomas-zimmerman (approved):
CC BY-SA is likely a better free license for documentation.

PR #323, opened 2025-10-14
serw14 (specs): Add 300Hz panel info & AC adapter option

jacobgkau:
Also corrects a minor typo found on the addw5 & bonw16 spec pages.


leviport:
It might be a bit awkward to work in, but I think it might be good to note that the 300Hz panel defaults to 240Hz unless the graphics mode in BIOS is set to discrete Nvidia only due to limitations of the iGPU.

leviport (approved):

PR #324, opened 2025-10-16
Add new desktop specs

jacobgkau:
This adds thelio-r5-n1, thelio-mira-r4-n3, and thelio-spark-r3-n3 specs to tech docs.


jacobgkau:
This is ready other than needing the part number for the 16GB RAM stick offered in Thelio Spark R3-N3.

PR #325, opened 2025-10-20
oryp13: Overview & repairs

jacobgkau:
This adds the external & internal overviews and the repairs manual for the Oryx Pro 13 (oryp13) to tech docs.


src/models/oryp13/img/cmos-battery.webp
leviport: This image appears in the CMOS battery section, but shows the speaker connector highlighted. Should the CMOS battery connector be highlighted instead?

PR #326, opened 2025-10-28
thelio-major-r5-n3 (specs): Add 1200W PSU info

jacobgkau:
Also added the existing info about the previous PSU split from the repairs page on the main specs page, since it seemed like it made sense to include it there now.


jacobgkau:
@leviport After consulting with the build team and @tonysystem76, we're apparently using the Core Shift II now and not the Core Reactor II. The Core Shift (II) is not on XPG's website yet. We're thinking it's better to use the correct product name and hold off on the link rather than using the closest-matching (but wrong) name & link.

PR #329, opened 2025-11-17
Changed the charger model name for the gaze17

AlexanderK-S76:
The charger listed for the gaze17 is the incorrect charger. The physical tip connector is the wrong size.

The actual two model numbers that are compatible with this laptop were updated.

A17-150P2A is what we'd ship if a new charger is purchased.


AlexanderK-S76:
Sager said the following when I just asked, hence why I proposed the change. Hopefully this helps.

>**What is the model number of the charger used for the gaze17 (3050 model)?**
>
>The 150W AC adapter model is either Chicony A15-150PIA or Chicony A17-150P2A
>
>**Is a Chicony A18-150P1A also compatible?**
>
>The Chicony A18-150P1A has 4.5x3.0mm tip. It does not fit the gaze17 5.5*2.5mm DC-in jack.

Issue #277, opened 2025-01-02
Issues opening the darp10(-b) & darp11 case


ahoneybun:
A few customers (two that I know so far personally) have had issues opening the case even with the tech-docs. This was the feedback from the latest:

"Even with all of the screws removed, my brother and I needed to pry at multiple points near the Ethernet port (the only spot where the bottom cover's edge was exposed for more than a tiny crack,) using the points of screwdrivers, in order to get enough of a grasp on the bottom cover to lift it off, without breaking anything, and it required a bit of a delicate touch.  The screwdriver-points also proved useful when disconnecting and re-connecting the CMOS battery (operating a bit like chopsticks or tiny fingers,) though unplugging and re-plugging the main battery was easy enough, once the cover was open."

I think for this model (since it splits in a different than expected location) folks are thinking it is right on the outside of the screws themselves:

https://tech-docs.system76.com/models/darp10/img/bottom-screws-16.webp

If we had photos where the split is (this can be true for other models going forward) that could help a bit.

ahoneybun:
ZD: https://system76.zendesk.com/agent/tickets/200484

I copied their message out but let me know if you need/want more context.

ahoneybun:
Some more feedback about this from Reddit:

https://www.reddit.com/r/System76/comments/1hxv4yk/one_month_followup_on_darter_pro/

"The System76 site has some great documents on this but one thing that I didn't understand was how to pull the bottom off once I had pulled all 11 screws off. You basically pull from just below the back hinges but I had a hard time at first. The system recognized the ram and new drive perfectly and I had no further issues there. "

leviport:
Do you think a side shot like this, that shows the back panel partially off, would get the message across?
![IMG_20250113_120559_835](https://github.com/user-attachments/assets/5a572321-f9c1-4475-b5f0-e1ae5c1b4b55)

We would obviously want to reshoot it with something better than my cell phone, but I'm just brainstorming how best to convey this concept. 

cc: @thomas-zimmerman

thomas-zimmerman:
I do think that either type of photo would help -- on brand new systems, the bottom panel is really hard to get started.

PR #333, opened 2025-12-09
darp10/11 (repairs): Show bottom panel & top case separation point

jacobgkau:
Closes https://github.com/system76/tech-docs/issues/277.

This angle seemed to make sense since the instructions say to start at the back, and it shows that the part wraps down and isn't just a flat piece on the bottom.


leviport (approved):
Looks great, and I think that should clarify where the edge of the bottom panel starts

AlexanderK-S76 (approved):
Changes look clear and leave no room for doubt on where the bottom panel pulls off from the rest of the chassis.

PR #350, opened 2026-04-08
meer9: Correct CU7 power adapter model number typo

leviport:
We noticed the model wasn't coming up as expected when searching it online, so we checked the actual adapter:

IMG_20260408_132845_202


jacobgkau (approved):
ADC017 must have been a typo, since while it does return some results, they appear to be much lower wattage than what the Meerkat would need.

PR #352, opened 2026-04-13
fix (various sections): Missing images & broken links

jacobgkau:
This fixes all missing images and broken links that @erinxocon flagged for us.

It also fixes a number of other broken intra-page anchor tag links in the messy combined Thelio Major B1/B2/R1/R2 section (which was ported from a pre-tech docs PDF by past team members), and removes the redundant YouTube links while keeping it printable (which I think was the point of duplicating them before).

For the Fn key and LED thumbnails (which are typically just copied from section to section), I converted the new ones from PNG to smaller files where appropriate, despite their tiny size to begin with. The ones being added as WebP here got larger instead of smaller when attempting to use AVIF.


erinxocon (approved):
LGTM!

PR #353, opened 2026-04-14
fix (ci): Track .avif files in Git LFS

jacobgkau:
This should've been done in https://github.com/system76/tech-docs/pull/347 when the first AVIF images were added to the repo, but it wasn't. As a result, 14 MB of AVIF files are now in the master branch's history.

To address this, the five commits before this (starting with 43e25e5) have been retroactively modified to place .avif files in LFS. The last commit adds the capitalized extension to the list to match the others.

A GitHub admin will need to force-push this branch to master. This will eliminate that 14 MB from the repo's history so it doesn't take that space up forever. (Otherwise, it would stick around in the history even if we converted the files to LFS pointers in the current commit.)

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,4 +1,5 @@
 *.avif filter=lfs diff=lfs merge=lfs -text
+*.AVIF filter=lfs diff=lfs merge=lfs -text
 *.bmp filter=lfs diff=lfs merge=lfs -text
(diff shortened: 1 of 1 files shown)


jacobgkau:
Still working on getting the timestamps right for the old commits.

leviport (approved):

jackpot51 (approved):

PR #359, opened 2026-05-05
bonw16: Add Core Ultra 9 290HX Plus CPU variant

### LINKED ISSUE #358: bonw16: New CPU should be added
The bonw16 will soon be updated to the [Core Ultra 290HX](https://www.intel.com/content/www/us/en/products/sku/245960/intel-core-ultra-9-processor-290hx-plus-36m-cache-up-to-5-50-ghz/specifications.html) CPU. This new CPU should be added to the list in specs.

leviport:
Closes #358


jacobgkau (approved):
We generally list higher-spec CPUs on top, so I swapped their position in the list. Looks good now.

PR #368, opened 2026-06-08
Erin/astro switch

erinxocon:
# Migrate docs site from mdbook to Astro Starlight

This PR replaces the mdbook setup with [Astro Starlight](https://starlight.astro.build/), giving us a modern, actively maintained docs framework with better theming, sidebar control, and a dev container for local development.

## Framework migration
- Replace `book.toml` with `astro.config.mjs`; add `package.json`, `pnpm-lock.yaml`, `pnpm-workspace.yaml`
- Tech stack: TypeScript, Node 24, pnpm 11

## Dev container
- Add `.devcontainer/Dockerfile` and `devcontainer.json` for reproducible local dev

## Content reorganization
- Move all docs under `src/content/docs/` to match Astro's content collections structure
- Rename root `README.md` files to `index.md` so it serves as a landing page, and rename section `README.md` files to `specs.md`
- Migrate first headings to frontmatter `title` fields

## Custom plugins & utilities
- **SUMMARY.md sidebar converter** — utility script that parses the existing `SUMMARY.md` into a Starlight sidebar config, preserving existing structure without a full content rewrite
- **Relative link plugin** — adds proper relative-link support for markdown files in content collections
- **Image wrapper plugin** — wraps every `` with a link to the original full-size file
- **Custom AVIF image service** (`avifImageService.mjs`)
- **Custom middleware** for ToC modifications

## Styling
- CSS overrides for System76 brand colors and sidebar label styles
- ToC width fix (`toc-width-fix.css`)
- Image border radius (`img-background.css`)
- Light/dark System76 logo assets added to `src/assets/img/`

## CI/CD
- Update `ci.yml`, `deploy-prod.yml`, and `deploy-staging.yml` for Astro build pipeline
- Add build caching step in production deploy
- Switch compression from gzip to zst
- Pin GitHub Action versions to commit hashes instead of tags to mitigate supply chain attacks
- Add Dependabot config for GitHub Actions and pnpm package updates

## Misc
- Rename entries titled "readme" to "specifications" in grouped sidebar sections
- Override static slugs for articles whose paths contain periods (Astro's docs loader drops them)
- Set `baseUrl` since the site is hosted at `/tech-docs`

PR #375, opened 2026-06-10
panp16: correct Ethernet from 2.5G to 1G

leviport:
The catalog was incorrect too, so that is also being corrected elsewhere.

### DIFF
--- a/src/models/panp16/README.md
+++ b/src/models/panp16/README.md
@@ -30,7 +30,7 @@ The System76 Pangolin is a laptop with the following specifications:
 - Memory
     - Up to 96GB (2x48GB) dual-channel DDR5 @ 5600 MHz
 - Networking
-    - 2.5-Gigabit Ethernet
+    - Gigabit Ethernet
     - M.2 PCIe WiFi 6E/Bluetooth 5.2
         - MediaTek MT7922A22M
 - Power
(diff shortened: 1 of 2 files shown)


jklgrasso:
This picture needs to be replaced as well:
https://system76.com/tech-docs/models/panp16/img/ports-right.avif

image

jklgrasso (dismissed):

leviport:
Last push broke the image's transparency. I fixed it differently this time, so it's closer to the original. Since I don't have the SVG, I had to get a little hacky, but I think it looks fine.

jacobgkau:
I have the SVG (I haven't offloaded that model into Dropbox yet, but I can also always do that at request), so no need to get hacky with it.

We're causing extra work for the impending software migration with every push at this point: https://github.com/system76/tech-docs/pull/368 I'd like to touch base with Erin and see if we're ready to go ahead and push that live this morning, or if we should merge this and make her rebase that again.

jacobgkau:
Erin's not available to go live with the switchover this morning, so she said to go ahead with our fix and she'll rebase when she has time again. I'll remake the image with the updated number.

jacobgkau (approved):
Resolution and chassis artifacting look better after re-exporting from the SVG with the updated text. Looks good locally.

PR #377, opened 2026-06-10
serw14: Update charger information for 290HX model

leviport:
I should've included this in my PR where I added the new CPU, but better late than never. The CPU update bumped the machine to a 280W charger:
IMG_20260610_142757_468

This can probably wait until after the Astro migration, if that makes things easier.

### DIFF
--- a/src/models/serw14/README.md
+++ b/src/models/serw14/README.md
@@ -51,12 +51,17 @@ The System76 Serval WS is a laptop with the following specifications:
         - Intel Wi-Fi 7 BE200
             - Bluetooth 5.4
 - Power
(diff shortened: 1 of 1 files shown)


jacobgkau (approved):
Slightly reformatted to better match how we've handled this for past laptops (e.g. galp5 & gaze17). This format more closely groups each variant's info together, which may reduce the effort to hunt around for the proper info for each model.

I'll check with Erin about merging, since she's working on rebasing her branch now.

jacobgkau:
Just made a slight commit message adjustment with the post-approval push.

PR #387, opened 2026-06-15
Optimize CI for Dependabot PRs and bump pnpm to 11.7.0

erinxocon:
Reworks the CI pipeline so Dependabot PRs skip steps they don't need, and updates pnpm.

Changes to `.github/workflows/ci.yml`:

- **Checkout**: for Dependabot PRs, use `fetch-depth: 1` and skip LFS, since these PRs only need lint and lockfile validation, not the full history required for the "last updated" date plugin or LFS assets used in the build.
- **Restore cache from S3**: skip for Dependabot, since Dependabot-triggered workflow runs don't have access to repository secrets and the step would fail.
- **Build**: skip for Dependabot, since it depends on the S3 cache to complete in a reasonable time and isn't needed to validate lint/lockfile changes.

Dependabot PRs still run checkout, install (`pnpm install --frozen-lockfile` — validates the lockfile is in sync), and format check, which covers the validation needed for dependency bumps.

Also bumps `packageManager` to `pnpm@11.7.0`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,8 +12,8 @@ jobs:
       - name: Checkout
         uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 #v6.0.3
         with:
-          fetch-depth: 0
-          lfs: true
+          fetch-depth: ${{ github.actor == 'dependabot[bot]' && 1 || 0 }}
+          lfs: ${{ github.actor != 'dependabot[bot]' }}
 
       - name: Setup pnpm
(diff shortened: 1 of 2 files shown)


jacobgkau:
The Dependabot PRs can be tested locally by just building their branches using the devcontainer commands, right?

jacobgkau (approved):

erinxocon:
> The Dependabot PRs can be tested locally by just building their branches using the devcontainer commands, right?

@jacobgkau Yes indeed! That is how i'm currently testing them!

PR #389, opened 2026-06-15
Add CODEOWNERS for dependency files and use cooldown instead of reviewers in Dependabot config

erinxocon:
**`.github/CODEOWNERS`** (new file)

Adds code ownership for `package.json`, `pnpm-lock.yaml`, and `pnpm-workspace.yaml`, assigning `@system76/web-engineering` as reviewers. This replaces the `reviewers` field previously set directly in `dependabot.yml` — GitHub now recommends CODEOWNERS for this since Dependabot's `reviewers` option only supports the legacy review-request flow and CODEOWNERS works more consistently across PR review requirements.

**`.github/dependabot.yml`**

- Removes `reviewers: ["system76/web-engineering"]` from both ecosystems (now handled via CODEOWNERS).
- Adds `cooldown.default-days: 1` to both ecosystems, so Dependabot won't open a PR for a new release until it's at least 1 day old.

This aligns Dependabot's PR-creation timing with our pnpm `minimumReleaseAge` policy (24h), preventing Dependabot from proposing package versions that CI will immediately reject for being too recently published — which was happening with `astro@6.4.7`.


jacobgkau (approved):
The web team previously did not need to allocate bandwidth (and could not become a bottleneck) for software (mdBook) updates in this repo. If the members of @system76/web-engineering are okay with needing to allocate the time to review this repo's dependency bumps, then this organizationally makes sense.

PR #390, opened 2026-06-15
thelio-major-r5-n4: External overveiw, internal overview, & repairs

jacobgkau:
This adds the Thelio Major R5-N4's service manual to tech docs. Some of the photos look a little rougher than the Mira R4-N4 section since the Major R5-N4 photo unit was a heavily used R&D unit.

I also corrected a number of problems I found in the Thelio Mira R4-N4 section's text while working on the R5-N4 section.

## Changing .md to .mdx for certain pages

This also corrects an issue with ampersands in image alt/title attributes. Astro was not rendering them properly:

Screenshot_2026-06-19_17-38-53

This seems to have to do with Astro's various (pre)processors for Markdown, HTML, etc. I'm not sure if it's technically a bug or not, but no amount of manual escaping or using alternative HTML codes for the & character was able to work around it. It appears the simplest way to correct it is just to use .mdx instead, which uses a more sophisticated preprocessor that doesn't have this problem:

Screenshot_2026-06-19_17-47-11

So I've gone ahead and converted all pages with an & inside an image alt/title attribute to .mdx (it's usually just a matter of renaming the file, as most valid Markdown syntax is also valid MDX syntax).

While working on that, I found a gap in our documentation regarding cleaning up after what's built by the devcontainer setup, so I also added that to the README.


README.md
```diff
@@ -33,3 +33,4 @@ The book can be viewed and edited on GitHub by going to
 - When finished editing, run `docker ps` to get the container's ID or name
 - To stop the container (for quicker setup next time), run `docker stop <container>`
 - To stop and remove the container (to start fresh next time), `docker rm -f <container>`
+    - Remove generated files outside of the container: `rm -r .astro .pnpm-store`
```
erinxocon, line 36: I would leave these unless you really need the space.

jacobgkau: When I changed a filename from `.mdx` back to `.md` during development, even completely removing the container and following the setup steps from scratch again was leaving me with an error on the affected page, as Astro remembered it had been an `.mdx` at some point and decided it wanted to always serve that.

The purpose of the "start fresh next time" steps are to start fresh. If someone wants to save time, they can use the "stop the container" method.

erinxocon:
I don't have any concerns about using MDX instead of MD. MDX should handle some edge cases better than MD like the above. As long as it builds, you can use whatever source format you want!

erinxocon (approved):
From a web point of view, it looks good. I'll update the build cache.

leviport (approved):
Looks great!

PR #407, opened 2026-07-13
Migrate to System76 brand fonts (Fira Sans, Roboto Slab, Ubuntu Mono)

erinxocon:
## Description
Replaces Starlight's default fonts with the brand typefaces used by the old Nuxt site's `@system76/design` package, loaded via Astro's built-in font provider.

- Adds `fonts` config in `astro.config.mjs` (Fira Sans for body, Roboto Slab for headings, Ubuntu Mono for code) using `fontProviders.fontsource()`
- Adds a custom `Head.astro` component to preload the body font and load the heading/mono fonts
- Sets `--sl-font`, `--sl-heading-font`, and `--sl-font-mono` in `variables.css` to the new families
- Fixes inline code and code block font size, which Starlight/Expressive Code default to smaller than body text (`--sl-text-code` / `--sl-text-code-sm` now match `--sl-text-body`) so code renders at the same size as surrounding text

### DIFF
--- a/astro.config.mjs
+++ b/astro.config.mjs
@@ -1,7 +1,7 @@
 // @ts-check
 import { satteri } from "@astrojs/markdown-satteri";
 import starlight from "@astrojs/starlight";
-import { defineConfig } from "astro/config";
+import { defineConfig, fontProviders } from "astro/config";
 
 import { satteriRelativeMarkdownLinks } from "@system76/satteri-relative-markdown-links";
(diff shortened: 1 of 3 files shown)


jackpot51 (dismissed):

jacobgkau (approved):
This decision was made (and approved) by the marketing team.

The change seems a little jarring to me, but I think it's just because it's different. I don't think it's particularly unreadable with Fira.

PR #415, opened 2026-08-05
oryp14: Specs

jacobgkau:
This adds some initial specs for the Oryx Pro 14 (oryp14).

The second commit corrects an outdated listing about Open Firmware in the lemp14 specs that was brought to light by a customer to the support team. (cc: @Alexander-Kendrick)


jacobgkau:
As I'll be off again next week, whoever reviews this for QA (probably @leviport) can feel free to make edits instead of simply requesting changes if anything needs to be fixed before it's merged.

leviport (approved):

PR #417, opened 2026-08-19
Update specs.md with BIOS details for processors

thomas-zimmerman:
Added BIOS version for Core Ultra and Core Series processors

### DIFF
--- a/src/content/docs/models/meer10/specs.md
+++ b/src/content/docs/models/meer10/specs.md
@@ -1,5 +1,5 @@
 ---
-title: Meerkat (meer10)
+title: Meerkat (meer10 & meer10-b)
 ---
 
 - [External Overview](./external-overview.md)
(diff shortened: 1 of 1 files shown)


src/content/docs/models/meer10/specs.md
```diff
@@ -16,6 +16,10 @@ The System76 Meerkat is a desktop with the following specifications:
         - [Core 7 240H](https://www.intel.com/content/www/us/en/products/sku/241653/intel-core-7-processor-240h-24m-cache-up-to-5-20-ghz/specifications.html)
         - [Core Ultra 5 225H](https://www.intel.com/content/www/us/en/products/sku/241749/intel-core-ultra-5-processor-225h-18m-cache-up-to-4-90-ghz/specifications.html)
         - [Core 3 100U](https://www.intel.com/content/www/us/en/products/sku/236776/intel-core-3-processor-100u-10m-cache-up-to-4-70-ghz/specifications.html)
+- BIOS
+    - Programmed with System76 firmware (non-open)
+    - Core Ultra (Arrow Lake) meer10
+    - Core Series (Raptor Lake) meer10-b
```
jacobgkau: The two lines just stating the CPU generation and their product code don't belong under the BIOS section without a BIOS-related reason to list them separately. I understand you made the connection that they use different firmware, but as is, it's not obvious and looks weird if someone didn't know that. (Also, "Ultra" vs. "Series" is not a valid comparison; the Ultra CPUs also belong to a Series. It's just Core vs. Core Ultra.)

This would be better presented in the CPU section above.

jacobgkau (changes requested):

src/content/docs/models/meer10/specs.md
jacobgkau: Taking a look at other pages for reference, the darp11 is a recent one that includes a `-b` variant: https://system76.com/tech-docs/models/darp11/specs In that case, we put it under BIOS because we were listing the motherboard part number, which differed between the two.

jacobgkau:
@thomas-zimmerman Do you still feel this gets the information you wanted across with how I've modified it?

leviport (dismissed):

leviport (approved):

thomas-zimmerman:
I am fine with the change! We just ran into not having the different versions documented anywhere and tech docs is where we go now.

jacobgkau (approved):

PR #418, opened 2026-08-20
Dependency updates

erinxocon:
Dependabot was having problems with something so I manually updated the dependencies. Tested locally on my computer and I reviewed the change log for astro to ensure nothing broke.

This should take care of the security warnings on the repo too.


jacobgkau (approved):
Built & ran fine locally.

PR #419, opened 2026-08-21
thelio-mira-r5-n4: Add usage note for USB4

jklgrasso:
Originally:

```
Using two high bandwidth devices (eGPU, USB4 dock, etc.) on both USB4 ports may limit port capability/device functionality. We recommend connecting only one high bandwidth USB4 device to the system at a time. Some USB4 devices may not work correctly if connected _after_ powering the system on.
```

@jacobgkau refined the original into bullet points :smile:


jacobgkau (approved):

jacobgkau:
Seems like @thomas-zimmerman is on-board, too, based on the Slack reaction.

PR #420, opened 2026-08-21
thelio-r5-n1 (specs): Remove ref to non-existent M.2 NVMe Gen 3 slot

thomas-zimmerman:
Removed mention of 1x M.2 (PCIe NVMe Gen 3) SSD from storage specifications that is not actually on the MB.

### DIFF
--- a/src/content/docs/models/thelio-r5-n1/specs.md
+++ b/src/content/docs/models/thelio-r5-n1/specs.md
@@ -61,7 +61,6 @@ The System76 Thelio is a desktop with the following specifications:
 - Storage
     - 1x M.2 (PCIe NVMe Gen 5) SSD
     - 1x M.2 (PCIe NVMe Gen 4) SSD
-    - 1x M.2 (PCIe NVMe Gen 3) SSD
     - 2x 2.5" SATA
 - USB
     - Back ports:


src/content/docs/models/thelio-r5-n1/specs.md
```diff
@@ -73,4 +72,4 @@ The System76 Thelio is a desktop with the following specifications:
         - 1x USB 3.2 Gen 1 Type-C
         - 1x USB 3.2 Gen 1 Type-A
 - Dimensions
-    - 32.7cm × 20.7cm × 29.1cm
\ No newline at end of file
+    - 32.7cm × 20.7cm × 29.1cm
```
jacobgkau, line 75: I'll fix this, but please try to figure out why your editor removed the newline at the end of this file and fix it.

jacobgkau (changes requested):

src/content/docs/models/thelio-r5-n1/specs.md
jacobgkau: Never mind, I guess there was no newline before (read the diff wrong).

jacobgkau (approved):

PR #421, opened 2026-08-21
fix (various/repairs): Use .md(x) extensions in relative links

jacobgkau:
Newly discovered Astro quirk: pages can be browsed with or without a trailing slash (or with `/index.html`), but relative links were breaking with the trailing slash since they weren't taking that extra directory layer into account (but _only_ on the live site for some reason; they worked fine when running it locally).

It appears that if we instead link to a source `.md` or `.mdx` file, Astro instead generates an absolute link that will always work. This also has the advantage of making the links work on GitHub (which is why we don't just put in absolute links ourselves).

### DIFF
--- a/src/content/docs/models/lemp14/repairs.md
+++ b/src/content/docs/models/lemp14/repairs.md
@@ -2,4 +2,4 @@
 title: Lemur Pro (Parts & Repairs)
 ---
 
-A service manual for the Lemur Pro 14 (lemp14) is not yet available. Please reference the service manual for the previous version, the [Lemur Pro 13 (lemp13)](/tech-docs/models/lemp13/repairs).
+A service manual for the Lemur Pro 14 (lemp14) is not yet available. Please reference the service manual for the previous version, the [Lemur Pro 13 (lemp13)](../lemp13/repairs.md).
--- a/src/content/docs/models/addp6/repairs.mdx
+++ b/src/content/docs/models/addp6/repairs.mdx
@@ -5,4 +5,4 @@ tableOfContents:
   maxHeadingLevel: 2
 ---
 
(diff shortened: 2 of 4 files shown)


src/content/docs/models/lemp14/repairs.md
```diff
@@ -2,4 +2,4 @@
 title: Lemur Pro (Parts & Repairs)
 ---
 
-A service manual for the Lemur Pro 14 (lemp14) is not yet available. Please reference the service manual for the previous version, the [Lemur Pro 13 (lemp13)](/tech-docs/models/lemp13/repairs).
+A service manual for the Lemur Pro 14 (lemp14) is not yet available. Please reference the service manual for the previous version, the [Lemur Pro 13 (lemp13)](../lemp13/repairs.md).
```
leviport, line 5: The other 3 changes use `.mdx`. Should this one match, or is it just `.md` instead?

jacobgkau: https://github.com/system76/tech-docs/blob/master/src/content/docs/models/lemp13/repairs.md is actually a `.md` file, while the other three are actually `.mdx` files. (We're only using `.mdx` for files that had an ampersand in an image description, since that shows up correctly if the file's `.mdx` but not if it's `.md`.)

I actually think the Astro parser doesn't care which one's used in a link (the sidebar generator definitely doesn't), but I'm trying to keep them in sync with what the actual filenames are anyway. And they need to use the actual file extensions to work when browsing the files on GitHub.

leviport: Interesting, got it.

leviport (approved):

PR #423, opened 2026-08-27
lemp13 (specs): Add 5600 MHz revision info

thomas-zimmerman:
Add `dmidecode` memory type from 5600 MHz version memory. Remove dead link to Samsung memory -- Samsung doesn't seem to have dedicated links for memory specs now. 

Add firmware model names to firmware section

### DIFF
--- a/src/content/docs/models/lemp13/specs.md
+++ b/src/content/docs/models/lemp13/specs.md
@@ -1,5 +1,5 @@
 ---
-title: Lemur Pro (lemp13)
+title: Lemur Pro (lemp13 & lemp13-b)
 ---
 
 - [External Overview](./external-overview.mdx)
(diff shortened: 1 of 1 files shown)


src/content/docs/models/lemp13/specs.md
```diff
@@ -18,6 +18,7 @@ The System76 Lemur Pro is a laptop with the following specifications:
     - [XMC XM25QU256C](https://www.xmcwh.com/en/site/product_con/224) flash chip
         - WSON-8 form factor
     - Programmed with [System76 Open Firmware](https://github.com/system76/firmware-open)
+        - `lemp13` or `lemp13-b` depending on keyboard
```
jacobgkau: This wasn't where it went in https://github.com/system76/tech-docs/pull/417, and I don't think it's where it should go here.

We already list the keyboard distinction (with pictures) at https://system76.com/tech-docs/models/lemp13/external-overview#keyboard--touchpad. Is that not sufficient? We can also add lemp13-b to the page header like I did in that other PR.

src/content/docs/models/lemp13/specs.md
```diff
@@ -38,8 +39,9 @@ The System76 Lemur Pro is a laptop with the following specifications:
         - 1x DisplayPort 1.4 over USB-C
 - Memory
     - Channel 0: 8-GB on-board DDR5
-        - 4x [Samsung K4RAH165VB-BCQK](https://semiconductor.samsung.com/dram/ddr/ddr5/k4rah165vb-bcqk/) (2GB each)
-    - Channel 1: 8-GB/16-GB/32-GB/48-GB DDR5 SO-DIMM @ 4800 MHz
+        - 4x Samsung K4RAH165VB-BCQK (2GB each) or
+        - 4x Samsung M425R1GB4PB0-CWMOD (2GB each)
```
jacobgkau: We should list the speeds with these two part numbers to distinguish them since that's the important difference.

It looks like the 5600MHz RAM was only offered in the lemp13-b (but not all lemp13-b units, only ones after July 2024)? @leviport does that sound right? If so, I would add the date (and model) info to the RAM section so people can have a chance at figuring out which RAM is in their unit based on that.

src/content/docs/models/lemp13/specs.md
jacobgkau: I've removed this new line and instead added `lemp13-b` to the page header.

src/content/docs/models/lemp13/specs.md
jacobgkau: I've shuffled this to separate the two RAM speeds by date.

leviport: That sounds correct, sorry for the delay. The `-b` variant distinguishes a keyboard layout variation and is unrelated to the RAM speed change which happened quite a bit later.

leviport (approved):

PR #425, opened 2026-08-31
addp6: External overview, internal overview, & repairs

jacobgkau:
This needs the included AC adapter model number. Should be complete aside from that.

### DIFF
--- a/src/content/docs/models/addp6/external-overview.mdx
+++ b/src/content/docs/models/addp6/external-overview.mdx
@@ -1,0 +1,76 @@
+---
+title: Adder Pro (External Overview)
+---
+
+## Left side:
+
+<span class="text-on-transparency">![Left Ports](./img/ports-left.avif)</span>
+
+## Right side:
+
+<span class="text-on-transparency">![Right Ports](./img/ports-right.avif)</span>
+
+## Front side:
+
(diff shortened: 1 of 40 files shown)


leviport:
Power adapter is a LITEON `PA-1231-26`

IMG_20260901_130933_686

edit: oh it's already added. Disregard.

src/content/docs/models/addp6/external-overview.mdx
```diff
@@ -0,0 +1,76 @@
...
+
+## Lid:
+
+![Lid](./img/lid.avif)
+
+## Materials:
+
+|Part                              |Material |Part Number    |
+|----------------------------------|---------|---------------|
+|LCD back cover (lid)              |Aluminum |6-39-L5501-022 |
```
leviport: These use the same magnesium/aluminum alloy that Lemurs use, so we should probably match how that's called out in lemp13

leviport (changes requested):
Just one thing, otherwise this looks great!

src/content/docs/models/addp6/external-overview.mdx
jacobgkau: Changed it from `Aluminum` to `Magnesium-aluminum alloy`.

leviport (approved):
🤘

PR #427, opened 2026-09-11
lemp14: External overview, internal overview, & repairs

jacobgkau:
This adds the external overview, internal overview, and repairs pages for the Lemur Pro 14 (lemp14).


src/content/docs/models/lemp14/specs.md
leviport: I just noticed the refresh rate here. It should be 165Hz for the 16" display. Guess I missed it in the initial specs PR.

src/content/docs/models/lemp14/external-overview.mdx
```diff
@@ -0,0 +1,133 @@
...
+|Key                       |Shortcut|Action                             |
+|--------------------------|--------|-----------------------------------|
+|![Fn-F4](./img/fn-f4.png) |Fn+F4   |Cycle keyboard backlight brightness|
+
+### 16" model (lemp14-b) only:
+
+|Key                            |Shortcut|Action                             |
+|-------------------------------|--------|-----------------------------------|
+|![Fn-*](./img/fn-star.png) |Fn+*    |Toggle keyboard backlight       |
+|![Fn-/](./img/fn-slash.png)|Fn+/    |Cycle keyboard color            |
```
leviport: lemp14-b's keyboard is just a single color, and this legend is not printed

leviport (changes requested):
Just a couple small things, but otherwise looks great!

src/content/docs/models/lemp14/external-overview.mdx
jacobgkau: I see, single-color but still uses numpad controls instead of a single key like the 14". I'll correct it.

leviport (approved):

PR #429, opened 2026-09-18
Update fn-star keyboard backlight symbol for many models

leviport:
I noticed the fn-star icon was incorrect on lemp14-b after I reviewed it. Then I poked around on a bunch of other models with the Fn + keypad * hotkey and realized that they were incorrect as well.

Going all the back to darp5 and darp6, the Fn + * symbol has been the circle with a horizontal line and "ray" dashes above it. The keyboard icon with "ray" dashes came from Fn + F4 on the tenkeyless keyboards, and must have been used for the fn-star icon by mistake.

This PR changes out the `fn-star.png` (and one `fn-star.webp` for oryp13) from the keyboard icon to the circle icon. There were a number that were already correct, so I left them untouched (oryp11 and oryp12, for example).


jacobgkau:
I think the reason the incorrect icon started being used was because we originally didn't have an image file for the correct one (or I made an assumption we didn't, at least). I actually went to check the lemp14's upstream manufacturer manual while working on that section (since it has both shortcuts on one page), but it seemed like they weren't including the keyboard legends in those manufacturer manuals anymore. I didn't realize we had the correct file to pull from any of our own older sections.

jacobgkau (approved):
Looks good when previewing these changes. I can't verify the accuracy of some of the older ones since older models didn't have a clear keyboard photo included, but the icon with the power symbol is definitely the norm for our numpad-including keyboards.

